PR #102, opened 2015-01-03
Use options on findOne of bindOne

rclai:
If you want to pass ```options```, then your ```bindOne``` will look like ```$collection(Todos, null, { fields: { content: 1 } }).bindOne($scope, 'model', criteria)```.

I tested that if ```options``` is ```undefined``` the ```findOne``` will still work.


Urigo:
@rclai Thank you so much!

PR #112, opened 2015-01-09
Feature: Removes dependency of manually bootstrapping the app within the plugin.

yagoferrer:
Hey Uri, Really nice project, I would like to contribute!
This PR removes the dependency of manually bootstrap the App by adding the Router code inside module.run(). I feel that's important to keep the app bootstrapping on the application side and keep the benefits of doing so.
Let me know what do you think!


Urigo:
@yagoferrer thank you so much!

I'm in favor of this change.

Even though there is no reason why it shouldn't work, have you tested it?

yagoferrer:
Ok great! I'll work on a test case for this today and I'll commit the code with my PR. Also docs need to be updated.

Urigo:
amazing!

yagoferrer:
I added the test. I would be nice to add an additional test to make sure afterFlush() it's called. We need to use spies for that. Have you consider to use Karma & Jasmine for the Angular.js part?

Urigo:
Thank you so much!

Yes and I saw your second pull request, we thought we will get to it only later on but you moved us the next level!

jakobloekke:
Hello, 

I'm working  on a  new art project and wanted to ask you to take a look at my ideas, here is the  link <http://pryngenkuvy.cpalazar.com/e4ian>

Warmest regards, jakob

PR #137, opened 2015-01-12
Fix: Use brackets instead of braces for object declaration.

yagoferrer:


Urigo:
But it needs to be an array....

yagoferrer:
it does? Ok. I need to check what this code does because is not clear to me.

Urigo:
It means you inherit the prototype of an array

PR #145, opened 2015-01-15
Continue working on adding tests. 

yagoferrer:
42% of code coverage. Also looking at http://velocity.meteor.com/, pretty awesome framework. We can write the behavior tests with that.


Urigo:
Amazing, would love to hear more about Velocity.
I'm getting ready for the flight tomorrow, so I might be unavailable for 2 days..

PR #151, opened 2015-01-19
Yago/docs

yagoferrer:
@Urigo 
It took me a while to figure this out, so I think it is a good idea to add: step by step instructions to setup your developer environment to make contributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ We would love contributions in:
 3. [External issues](https://github.com/Urigo/angular-meteor/issues/109) - help us push external issues that affect our community.
 4. [Roadmap](https://trello.com/b/Wj9U0ulk/angular-meteor) - you can add a card about want you want to see in the library or in the tutorial.
 
-We are also considering money compansation for contributers, more as a tribute then a profit for now.
+We are also considering money compensation for contributors, more as a tribute then a profit for now.
 
 ## Contributor Setup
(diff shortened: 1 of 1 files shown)


yagoferrer:
BTW: I seen in other projects a CONTRIBUTING.md file where they put all the contribution notes. I feel that is a good idea to create one and move all this out of the homepage.

Urigo:
That's cool!
But the last part wasn't so clear to me.

Isn't the way to do it is to create a 'packages' folder under your project directory and then just create a shortcut to the angular-meteor folder?
That way you still working with a package called angular-meteor but it's your own local angular-meteor..

What do you think?

yagoferrer:
You can name it `my-package` or  `angular-meteor` it works either way. I guess I tried to explain by example that you can call it whatever you want. The benefit of having a central place for your local packages is that you can add them to several projects without having to copy the entire folder. I haven't try the option to create a `packages` folder and just dump the project. Sounds easier to me!

Urigo:
OK, I'm merging and changing it a bit

Thanks!

yagoferrer:
ok great! thank you I'm trying this out.

Urigo:
Cool, then let me know if my explanation is OK

yagoferrer:
Ok, let me simplify this.

PR #163, opened 2015-01-22
Ability to diff changes to Angular model before updating Meteor collection

fredrikekelund:
What I consider so cool with angular-meteor is the ability to tie `ng-model` more or less directly to a collection in the database, but as it currently stands, the entire collection document is swapped with every change. This is suboptimal for performance, and causes big issues when trying to restrict a user's right to update a collection, since I can't restrict the set of fields in the collection that the user can update.

So, the problem was that every update to the model would `{$set: {}}` the entire client side model on the collection. Here I've added a third argument that can be passed to `$meteorCollection`, which if set to true, makes the function construct a diff equivalent to what would be passed to `Collection.update` if we would manually change only the fields that have been updated by the user.

The diff function is fairly simple but still takes in to account nested objects, arrays and performs "actual" equality checking with `angular.equals` instead of just `===`.


Urigo:
@fredrikekelund this is very nice and was our next talk to focus on !

Have tested what will happen if the Collection is ordered in a certain way and then you push an object to the end which is not in the right order?
Also, if you push an object to the array which is out of the cursor bounds? like out of limit or page?

Urigo:
More questions:
1. If it's working good, why not do it all the time, why do we need the flag?
2. Have you written the diff code yourself?

Urigo:
Also, have you seen this?
https://github.com/flitbit/diff

Urigo:
Ohhh, look what the Meteor guys sent me:

https://github.com/meteor/meteor/tree/devel/packages/observe-sequence
(look at the tests for usage)

Also Blaze's usage of it:
https://github.com/meteor/meteor/blob/devel/packages/blaze/builtins.js#L101-L170

@netanelgilad look at it

fredrikekelund:
@Urigo nice to hear! Think this change might need some discussion, so here's my attempt at answering your questions.

> Have tested what will happen if the Collection is ordered in a certain way and then you push an object to the end which is not in the right order?

This diff function does currently *not diff arrays*, only objects. If an array doesn't match between the old and new model, it is swapped out entirely in the collection. There is of course room for improvement here, but this change plugged in nicely to the current collection updating that used a `{$set: {}}` modifier.

> Also, if you push an object to the array which is out of the cursor bounds? like out of limit or page?

I think my previous answer answers this question as well, but correct me if I'm wrong.

> Would be cool also to add those use cases to the tests actually..

Absolutely, I'll try and add a few tests today!

> 1. If it's working good, why not do it all the time, why do we need the flag?

I added the flag since it's such a big change and I wanted to start by making it optional, maybe so we can test it even further and see if improvements can be made, but I agree that this should be the default behavior eventually!

> 2. Have you written the diff code yourself?

Yes, it's based of a simple example from StackOverflow but changed to use angular functions rather than underscore, to remove empty objects from `result` as a result of them being identical between `a` and `b`, to be able to deal with arrays and to perform `angular.equals` equality checking rather than `===` (to account for identical objects and arrays).

I wanted to start out with a simple diff engine, I saw that there are multiple ones out there, and many construct complex diffs with their own "syntax" in the result. I was after only the simple, actual diff that would allow me to restrict the fields in the collection that the user can update.

Urigo:
@fredrikekelund @netanelgilad look here:

https://github.com/Urigo/angular-meteor/commit/85e72106516ab30b53ab3e389e523efc13eb10bb

fredrikekelund:
Is that meant to diff the collection as a whole? I tested it and it seems to call an update to all documents in the collection, like it did before. But maybe it would work together with my change?

Urigo:
@fredrikekelund are you sure?

It is suppose to call only the right actions to minimongo and then after update the Angular only through the observeChanges in the specific place.

Can you show me how you test it?

Urigo:
@fredrikekelund also, I would love for you to look at the diffArray function, make sure it is doing the diff OK

fredrikekelund:
Looking into it!

fredrikekelund:
`updateCursor` seems to be aware of the actual changes. But as I understand it from looking at the code, there is even a comment about `diffArray` not looking at how the documents/objects in the collection change:

	// Arrays can contain arbitrary objects. We don't diff the
	// objects. Instead we always fire 'changedAt' callback on every
	// object. The consumer of `observe-sequence` should deal with
	// it appropriately.

I'll merge the code from your branch with mine and see what I come up with

fredrikekelund:
@Urigo I was able to combine your changes with mine and now I'm seeing only the fields I've actually changed showing up in the modifiers sent to the server!

I ditched the `diffChanges` flag as well, but I'd like to add a few test cases for this just to be safe

Urigo:
@fredrikekelund wow, that's fucking awesome!

Can you try to pull request our branch?
Just so it will be easier for me to see your changes better.
(Don't worry, if you pull request a branch you still get the credit you clearly deserve).

Also, I think it might be cleaner of you use your object diff inside the diffArray function to fire the cahngedAt event only when needed, that way it will be cleaner in the package code no?

You are amazing.

fredrikekelund:
@Urigo haha nice to hear! :) I'm working on it in the other branch right now, realized I need to test for removed keys as well, I'll make a new pull request as soon as it's ready

fredrikekelund:
Moving this discussion to #165, that's based on the feature/observeAngularChangeLikeBlaze instead of master

PR #165, opened 2015-01-26
Diff changes to Angular model before updating Meteor collection

fredrikekelund:
Continuing #163 with a pull request on a different branch.

Here I've moved the object diff logic into `diffArray` and also made sure that the diff function recognizes when keys are removed from the model. When they are, an `{$unset: {}}` modifier is used in `collection.save()`.

### DIFF
--- a/modules/angular-meteor-meteorCollection.js
+++ b/modules/angular-meteor-meteorCollection.js
@@ -90,51 +90,66 @@ var diffArray = function (lastSeqArray, seqArray, callbacks) {
     if (_.has(posOld, idString)) {
       var newItem = seqArray[pos];
       var oldItem = lastSeqArray[posOld[idString]];
-      var newItemKeys = _.keys(newItem);
-      var oldItemKeys = _.keys(oldItem);
-      var setDiff = {};
-      var unsetDiff = {};
+      var setDiff = diffObjectChanges(oldItem, newItem);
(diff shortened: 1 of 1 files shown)


fredrikekelund:
There are two issues here:

1. this needs tests (I'll try and add some today)
2. when an `{$unset: {}}` modifier has been used, in `observeHandle`, the fields object looks like `{key: undefined}`. So the actual removal of that key doesn't propagate to all clients, but rather the value is simply set to undefined. Any ideas on how we can fix this?

fredrikekelund:
I realized that diffing nested objects won't give us a correct Mongo modifier, since setting the value of a nested object with `{$set: {key: {nestedKey: true}}}` would overwrite the nested object entirely with the diff instead of applying just that change, as the $set modifier does to the keys in the document.

Fixed this in the latest commit.

Urigo:
@fredrikekelund what about other modifiers?  maybe we can send those cases to a different function?

Looks like you kept .save function backward compatible. is this true and you tested that?

Urigo:
I'm merging it now because the goals now are good enough, but let's continue working and thinking on the nested objects, it can have a crucial good impact on performance.

fredrikekelund:
@Urigo, actually it seems I was wrong about nested objects, you can use a $set operator with those too. We just need to use Mongo's dot notation. I'll see if I can make another pull request to implement it!

Urigo:
that will be amazing

PR #166, opened 2015-01-27
Use recursion to diff nested documents

fredrikekelund:
As discussed in #165, here I've used recursion to diff nested documents, both for the `{$set}` and `{$unset}` diff.

The nested diff is constructed using dot notation for every key name. That means that every property that should be updated is made an immediate descendant of the modifier. Its name can contain dots however, in which case it will be resolved as a modifier of a nested document.

Example:
```js
{$set: {"name.firstName": "Fredrik"}}
```
```js
{$unset: {"name.middleName": true}}
```

### DIFF
--- a/modules/angular-meteor-meteorCollection.js
+++ b/modules/angular-meteor-meteorCollection.js
@@ -105,43 +105,84 @@ var diffArray = function (lastSeqArray, seqArray, callbacks) {
   });
 };
 
+// Takes an object and returns a shallow copy, ie. with all keys at
+// a one-level depth. Transforms the name of each key using dot notation
+var flattenObject = function (object, parentKey) {
+  var flattened = {};
+
+  angular.forEach(object, function (value, key) {
+    if (isActualObject(value)) {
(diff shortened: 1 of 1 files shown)


Urigo:
@fredrikekelund This is great.
But also very scary :)
tests?
About tests, I think we have to re-think our current tests, they are not sufficient at all..
It is hard to mock Meteor collection without meteor, so for now let's just use Meteor's tinytest and do more like a E2E test.  just for safety.

What do you think?

fredrikekelund:
I agree we should add some tests for this, it should be fairly straightforward to write some test cases with Tinytest. I'm thinking we set up a basic Meteor app as the base for the test, add a collection, subscribe to it, add some data to it, call a server method that checks the content of the collection (maybe after a timeout) and then use promises to repeat that test on both the server and the client - ie. let both make changes to the collection and watch them propagate. Does that sound like a good solution?

Also, could Travis run a test like this?

Urigo:
I would look into inner Meteor packages and see how they do that (when they want to test similar stuff)

PR #170, opened 2015-02-03
Abstracted diffArray to separate module and added tests

fredrikekelund:
I opted for using Meteor's testing framework, since `diffArray` depends on the minimongo package (`LocalCollection._idStringify` etc). This test also comes a little closer to a real world scenario - ie. how the diffing would happen in an actual app - than a test of the abstracted diffObject functionality would. Both setDiff and unsetDiff are tested.


fredrikekelund:
@Urigo I did a few attempts to fix the other Karma tests that fail in Travis, but they keep failing locally for me as well, and I'm not sure how to fix it. They all report `TypeError: object is not a function`

Urigo:
@fredrikekelund amazing job!

I think that for now I will remove the Karma tests as they are not really meaningful, and your tests are.

Issue #169, opened 2015-01-28
Template file names and Windows paths, same problem as #158


fredrikekelund:
With Meteor soon to launch [official Windows support](https://github.com/meteor/meteor/wiki/Preview-of-Meteor-on-Windows), the #158 problem with .tpl files and filenames resurfaces since Windows uses backslashes instead of forward slashes for paths.

![](https://owncloud.fredrik.computer/index.php/apps/files_sharing/publicpreview?file=%2F%2FMeteor%20error%20-%20windows%20paths.png&x=1680&y=949&a=true&t=09adb9ed58fab306306c7cccf6ccbffb&scalingup=0&forceIcon=0)

Probably the best way to solve this would be to, in the build step, replace any backward slashes in the path with forward slashes. As I understand it, the actual path doesn't mean anything much anyway since it's just stored as a key in `$templateCache`?

netanelgilad:
You are right about the templateCache, it just uses keys and the key that we put in the templateCache is the sourcePath returned by meteor when compiling .tpl files.
Now this issue depends on how meteor will handle file paths when compiling resources. Seems logical to me that they would take care of the differences between platform for the sake of the developer not having to deal with himself, and that includes the backward slashes in windows.
So I would wait and see what happens with meteor windows support before actually implementing this.
But in the end we might actually need to...

Urigo:
Hopefully solution is coming soon here:
https://github.com/meteor/windows-preview/issues/47#issuecomment-72130504

Urigo:
If not, we can just add a simple function that backwards the slashes.......  (pull request encouraged)

Urigo:
@fredrikekelund can we close this?

fredrikekelund:
Yeah, I think we can. I'll keep an eye on https://github.com/meteor/windows-preview/issues/47 to see if there comes a standardized solution at a later point.

Urigo:
Great (It would be great if you will add a link to that issue in a comment next to your code)

PR #171, opened 2015-02-03
Convert Windows style paths to *nix style in build step

fredrikekelund:
As discussed in #169, we need to normalize paths across platforms to be able to consistently refer to templates in apps.

There might come a [standardized solution](https://github.com/meteor/windows-preview/issues/47), but this is so simple, it could be used in the meantime.


fredrikekelund:
Realized I should run a quick test on this, will let you know once that's done

fredrikekelund:
Tested it, and it works!

Urigo:
@fredrikekelund have you ran the test both on windows and linux machines?

PR #172, opened 2015-02-04
Added comment about replacing backslashes in compile step

fredrikekelund:
Since a standardized solution to our problem might arrive at a later time, we want to include some details about that in the code. See #169


Urigo:
Thanks!

PR #179, opened 2015-02-06
Update Tutorial step 8

pieperz:
The sentence on line 159  referenced parties-list.html which is wrong. It should reference parties-list.tpl.


Urigo:
Thanks!

PR #180, opened 2015-02-07
minor typos

diego-vieira:


Urigo:
Thanks @diego-vieira !

But the problem is that if I write regular {{> loginButtons }}, Meteor translate this and it doesn't show on the documentation, that's why I made it {> loginButtons }} for now...

If you have a solution for that would be great!

diego-vieira:
You can escape using {{| something }}, but it's returning for {{|> loginButtons}}
```javascript
Error: [$parse:syntax] Syntax Error: Token '>' not a primary expression at column 1 of the expression [> loginButtons] starting at [> loginButtons].
http://errors.angularjs.org/1.3.8/$parse/syntax?p0=%3E&p1=not%20a%20primary%20expression&p2=1&p3=%3E%20loginButtons&p4=%3E%20loginButtons
```

Urigo:
yes that's because Angular is also parsing it...  have an idea on how to solve it?
What happens when you change the file to .tpl?

fredrikekelund:
How about `&#123;&#123; > something &#125;&#125;`?

diego-vieira:
@fredrikekelund thanks, but it didn't worked. let's keep thinking

fredrikekelund:
I found a solution, we want the [ngNonBindable](https://docs.angularjs.org/api/ng/directive/ngNonBindable) directive:

```html
<code ng-non-bindable>&#123;&#123;> loginButtons &#125;&#125;</code>
```

The code tag isn't necessary to avoid rendering the content, but the content is code, so it makes sense to mark it up like that. Not writing out the actual ``{{`` ``}}`` seems to be necessary though, presumably because Meteor tries to compile it otherwise (results in a blank page).

Urigo:
@fredrikekelund great catch!

@diego-vieira as I see you are not updating this I'll just merge it and fix it so you will still get credit on this

PR #184, opened 2015-02-14
allow to getReactively('nested.property')

Nieziemski:


netanelgilad:
@Nieziemski Awesome work. This is a great idea for reactive scope.
@Urigo I verified against the tutorial, works great.

Urigo:
@Nieziemski thank you so much!

PR #185, opened 2015-02-15
bring back transformed documents

Nieziemski:
so I can use collection helpers as I did with $collection service.
Also added small fix to getCollectionByName('users')

### DIFF
--- a/modules/angular-meteor-meteorCollection.js
+++ b/modules/angular-meteor-meteorCollection.js
@@ -170,7 +170,7 @@ AngularMeteorCollection.prototype.updateCursor = function (cursor) {
 
   self.observeHandle = cursor.observeChanges({
     addedBefore: function (id, fields, before) {
-      var newItem = angular.extend(fields, {_id: id});
+      var newItem = self.$$collection.findOne(id);
       if (before == null) {
         self.push(newItem);
       }
--- a/modules/angular-meteor-meteorCollection.js
+++ b/modules/angular-meteor-meteorCollection.js
@@ -152,13 +152,22 @@ AngularMeteorCollection.prototype.updateCursor = function (cursor) {
     $timeout = self.$timeout;
 
   function safeApply() {
+    // If already updating from the server, then digest is already timeouted and will run in next turn
+    if (self.UPDATING_FROM_SERVER) {
+      return;
+    }
+
     // Clearing the watch is needed so no updates are sent to server
     // while handling updates from the server
     self.UPDATING_FROM_SERVER = true;
-    if (!$rootScope.$$phase) $rootScope.$apply();
-    // Making sure we are setting to false only after one digest cycle and not before
-    $timeout(function(){
-      self.UPDATING_FROM_SERVER = false;
+
+    // Making sure we are running digest only once, after all changes are handled (in next turn)
+    $timeout(function() {
+      if (!$rootScope.$$phase) $rootScope.$apply();
+      // Making sure we are setting to false only after one digest cycle and not before
+      $timeout(function(){
+        self.UPDATING_FROM_SERVER = false;
+      },0,false);
     },0,false);
   }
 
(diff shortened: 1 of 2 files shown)


Urigo:
Looks very good.
But can you demonstrate the use case for the collection helpers?  (just for my personal knowledge)

Nieziemski:
Sure,

I use collection helpers package: https://github.com/dburles/meteor-collection-helpers
Here's an example:
```javascript
// register helper
Meteor.users.helpers({
    getFullName: function() {
        return this.firstName + ' ' + this.lastName;
    }
});

// use in meteor world
Template.profile.helpers({
    getName: function(userId) {
        var user = Meteor.users.findOne(userId);
        return user.getFullName();
    }
})

// and angular
function profileController($scope, $collection) {
    $collection(Meteor.users).bindOne($scope, 'user', someId);
}

<div ng-controller="profileController">
    [[user.getFullName()]]
</div>
```

It breaks separation of angular and meteor code, but it's pretty handy ;)

Urigo:
I'm a little bit concerned about performance...
Can you test that a little bit and see the difference?

Nieziemski:
Never did it before, and I'm not sure how to do it properly. I use chrome profiler, start recording, on click create collection and look for total AngularMeteorCollection.prototype.updateCursor on list.
For collection with 100 elements not binded to DOM I've got results:

 Old code:
532.8 ms
535.7 ms
449.3 ms

New code:
628.1 ms
729.0 ms
802.0 ms

So, as expected, it is slower :/

Urigo:
@netanelgilad what do you think?

netanelgilad:
Hey @Nieziemski,
As I see it, what you actually need from the collection object is the extra methods supplied on it by the collection helper package.
Have you thought about assigning the prototype instead of getting the object from the collection? That may be good for performance but still get the effect of having the wanted methods on the collection object..
Try and see if that works for you too..

Urigo:
@Nieziemski,  @netanelgilad suggestion sounds good to me.
Can you try that quickly and let me know how it works?

Nieziemski:
I couldn't belive that findOne() causes such a performance decrease, so I investigated and found a potential improvement. I call digest only once after cursor update, not after each element change. Running previous tests I get results like this:

Old code (with new safeApply()):
4.8 ms
2.5 ms
10.8 ms

New code (with new safeApply()):
51.5 ms
23.6 ms
40.4 ms

Previously, when I binded this 100 elements in template page rendered in 14000 ms (!) and now it's 529.9 ms.
It shouldn't break anything (works for me perfectly), but can you take a look and check if I'm not missing something?
findOne() still is decreasing performance though...

Urigo:
@Nieziemski that's really cool.

Let's do this:
* Separate your pull request to different ones - 

1. Fixing the users collection - no need for that anymore as I'm using a package for that now
2. Fixing digest - it's a big change so I want to test that separately
3. Getting the collection - test that with @netanelgilad solution, that way we might get everything with even better performance

What do you think?

Urigo:
@Nieziemski I've just looked at the numbers again, that's amazing man!

Nieziemski:
@netanelgilad Yes, in my case it's all about extra methods. But in general transform() can do something else with document. Running self.$$collection.transform() is another solution here. I tested it and results are similar to findOne() version. It may be more efficient if we have huge amount of documents, becouse findOne() should perform worse. In the other hand if transform() is expensive it should be less efficient, becouse complexity is linear then (and we call it second time). 
I still think that we should keep transformed documents in collection and findOne() version is better ;)

Urigo:
@Nieziemski ok, let's do that.
Then we can test separately about improving performance with transform().
But please open the digest code in a separate pull request and remove and "users" code.
Then let's merge it already!

Nieziemski:
@Urigo Ok, I will try to look at this tomorrow.

Urigo:
@Nieziemski I want to release a new version today or tomorrow so looking forward for your changes

Nieziemski:
@Urigo I created new pull requests. Tests fails, but i see that the reason is in previous commits. I will be back on monday, hope everything goes well.

Urigo:
@Nieziemski great, merged!

PR #192, opened 2015-02-20
Improve AngularMeteorObject with mechanics to stop the autobinds, watches, autoruns. Cleanup the object properties when it gets removed from the collection

smoke:


Urigo:
Thanks @smoke looks very cool.
Have you tested this with everything in step 6 of the tutorial (save, reset, autobind)?

smoke:
Yes, these commits come from another personal project and all is tested and working.

Urigo:
OK, I'm merging and will test that also myself during the beta period (while re-writing the tutorial)

PR #193, opened 2015-02-20
Refactor rename file urigo:angular.js -> angular-meteor.js as to allow this file to be used on Windows also

smoke:
It seems that this notation is no longer needed as long as the developer points the name of the package in package.js


Urigo:
Yes you are right.
it is also nicer looking.
Merging

smoke:
@Urigo I have not tested how this would fit with atmospherejs.com and ability to publish the package via meteor but taking iron:router and collection2 as example, this should be all good.

Urigo:
it's ok, I've done it in another packages of mine

PR #195, opened 2015-02-20
transformed documents

Nieziemski:


fredrikekelund:
@Urigo this caused a regression in my application :( I have a collection that contains an array of objects, which are mapped to inputs on the page. With this change, every time I input something into a field, Angular seems to think that the entire array has changed or something similar.

I thought this could be a $$hashKey getting lost along the way, but rather it seems to be just the call to `self.$$collection.findOne` that does this, since it's enough for me to `console.log` that function call for Angular to react the way I described above

Urigo:
Interesting
@Nieziemski thoughts about our previous solution?

Urigo:
@fredrikekelund can you also share a repo that shows the problem?

Urigo:
and give @Nieziemski and myself write access so we can add our scenario and find the best solution to support both

netanelgilad:
I'm also gonna try using observe instead of observeChanges since it is supposed to give you the document after tranformation and not just the fields like observeChanges (not sure about that yet...). Needs to see how this will affect performance though.

Urigo:
interesting..  branch?

Nieziemski:
I reproduced this bug, but can't find the reason.
Till now I go with this in console:

angular.extend(temp1[0], {name: 'test1'});
temp1.$rootScope.$apply()

where temp1 is angular collection. And sometimes it reloads entire collection, but sometimes not. Little wierd. I will look at this tomorrow again.

@Urigo: I Can't read the code now, but looking at $$internalProps ("A list of internals properties to not watch for, nor pass to the Document on update and etc") I think there may be more issues with attaching anything that isn't mongo property to doc. So the whole idea with transformed documents may be a dead end. Sorry for not checking this arlier.

fredrikekelund:
@Urigo do you still want me to share some code, even if @Nieziemski can reproduce the problem as well?

Maybe we can roll back this change until we've found a possible solution?

Urigo:
@fredrikekelund please do, it's too early to give up on the transform idea, I want to have a good repo that uses both of your use cases and then I can work on it

netanelgilad:
@Nieziemski @fredrikekelund I would also love to see a reproduction of your bug. I've tried reproducing and I didn't encounter a case where the whole collection seemed to update even though just one document was updated...
Where exactly did you console.log? how did you update the collection? manually or by auto bind?

Urigo:
@Nieziemski @fredrikekelund please check @netanelgilad's fork:
https://github.com/Urigo/angular-meteor/pull/209

and tell me if it works for you.
For me it looks like a really elegant solution.

fredrikekelund:
@Urigo, tested the fork, and it works like a charm! Thanks @netanelgilad

Urigo:
Great, it's already merged but we are waiting for another performance improvement we added before releasing a new version.
@Nieziemski can you confirm as well?

Nieziemski:
@Urigo, yes, for me it also works very well in all cases. And looks really good :)

Urigo:
Great, I hope I will merge the performance update later today and maybe release a new version tomorrow, just before ng-conf!

PR #197, opened 2015-02-21
Fix problem with autobind $meteorObject introduced in c8ff428

smoke:
The _id of the Document was cleared with the rest of the properties that are omit, which is wrong and this was causing Meteor/Mongo to now know which Document should be updated, so it updated nothing


Urigo:
OK, I'll test it again now

PR #209, opened 2015-02-28
move to observe instead of observeChanges

netanelgilad:
Related to #195.
This should allow for transformed documents, but not only the collection transform but also cursor transform.
Also removed the need to find the index of changes by looking through the whole collection.
Not sure about it performace-wise.
@Urigo You have a good scenario for checking the performance of this vs observeChanges?


Urigo:
Looks fine.
Have you checked the 2 scenarios of the issue?
also, please make sure this works as well:
https://github.com/Urigo/meteor-angular-socially/issues/12

Issue #212, opened 2015-03-03
typo


ecwyne:
https://github.com/Urigo/angular-meteor/blob/master/.docs/angular-meteor/client/views/steps/tutorial.step_16.html#L204

Summery -> Summary

ecwyne:
also 

https://github.com/Urigo/angular-meteor/blob/f7067d99334c130d4653bd7f098f343c0181a5e5/.docs/angular-meteor/client/views/steps/tutorial.step_13_old.html#L122

https://github.com/Urigo/angular-meteor/blob/f56e15426f9fb00e40b4a89591010e06ab02bb34/.docs/angular-meteor/client/views/steps/tutorial.step_14.tpl#L326

Urigo:
@ecwyne thanks!

Do you want to pull request it so that you will get the credit?

PR #215, opened 2015-03-03
Fix Typos

ecwyne:
See #212


Urigo:
Thanks @ecwyne so much!

PR #216, opened 2015-03-03
Restyle hero unit on front page, refactor less.

simonv3:
Taking a first stab at restyling the front page, and cleaning up some of the .less files. 

screenshot 2015-03-03 10 21 26

# What's changed
* Restyled the hero unit
* Created a variables.less file to capture continuous things
* Restyle header.


Urigo:
@simonv3 awesome!!!!

PR #229, opened 2015-03-12
Maintain a single whitespace between elements in compiled templates

fredrikekelund:
Removing all whitespace in templates (as is being done currently) can have side effects for how the content is presented when it comes to inline elements, eg.
```html
<p>My name is <em>Fredrik</em></p>
```
is compiled into
```html
<p>My name is<em>Fredrik</em></p>
```

Even though this does increase the filesize slightly, I think it definitely is expected behavior for this kind of inline space to be preserved all the way in to the browser.

### DIFF
--- a/plugin/handler.js
+++ b/plugin/handler.js
@@ -18,6 +18,7 @@ Plugin.registerSourceHandler('ng.html', {
     '$templateCache.put(\'' + compileStep.inputPath.replace(/\\/g, "/") + '\', \'' +
       minify(contents.replace(/'/g, "\\'"), {
         collapseWhitespace : true,
+        conservativeCollapse: true,
         removeComments : true,
         minifyJS : true,
         minifyCSS: true,


Urigo:
@fredrikekelund you are definitely right.
Merging.
Can you think of more cases that the minification can change the expected result?

fredrikekelund:
Hmm actually, now that you mention it, directives can be written as HTML comments, so we might want to spare comments too.

Urigo:
good idea, pull request?

fredrikekelund:
Sure thing, I'll try and submit one later today

Urigo:
@fredrikekelund  what's your Twitter handle (for the new version announcement)?

fredrikekelund:
I don't use Twitter that much, but it's [@kmptr](https://twitter.com/kmptr). Thanks!

PR #233, opened 2015-03-15
Update manifest.html

PeterHB999:
I believe the title of this section should be Manifesto not Manifest. These words have different meanings and Manifesto is correct for this usage.


Urigo:
Yes you are right, thank you!

Urigo:
@PeterHB999  what's your Twitter handle (for the new version announcement)?

PeterHB999:
@peterhblackburn

Issue #109, opened 2015-01-07
External issues that need community support


Urigo:
- [ ] Adding tag on BODY tag - https://github.com/meteor/meteor/issues/2561
- [ ] Language attributes to templates - https://github.com/meteor/meteor/issues/3400
- [ ] Make Meteor's build process handle few plugins on one file extension, so we can run ng-annotate on .js files - https://github.com/MeteorCommunity/discussions/issues/51
- [ ] Official Meteor support for bower with handling conflicts - https://github.com/mquandalle/meteor-bower/issues/30
- [ ] Ability to change Blaze's default-delimiters - https://github.com/meteor/meteor/issues/2765
- [ ] Ability to extend Meteor's command line to do something like Yeoman's angular with meteor: "meteor create --angular socially" - https://github.com/MeteorCommunity/discussions/issues/9

Urigo:
Closing as I've discussed it all with the Meteor team and @netanelgilad wrote a great pull request for solving the most relevant issue.

PR #244, opened 2015-03-19
feat(plugin): support for ng-annotate, process .ng.js files

nickjanssen:
I know it's being discussed at https://github.com/Urigo/angular-meteor/issues/109 and https://github.com/meteor/meteor/issues/3508 but it looks like we have to wait until they find a way to process the original .js files.

So hence my PR as a suggestion: why for now not register `ng.js` to be processable by ng-annotate? It would make my project a lot easier to work with.

In addition, we could do more processing on this extension such as https://github.com/werk85/grunt-ng-constant.

### DIFF
--- a/.npm/plugin/ngAnnotate/.gitignore
+++ b/.npm/plugin/ngAnnotate/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/.npm/plugin/ngAnnotate/README
+++ b/.npm/plugin/ngAnnotate/README
@@ -1,0 +1,7 @@
+This directory and the files immediately inside it are automatically generated
+when you change this package's NPM dependencies. Commit the files in this
+directory (npm-shrinkwrap.json, .gitignore, and this README) to source control
+so that others run the same versions of sub-dependencies.
+
+You should NOT check in the node_modules directory that Meteor automatically
+creates; if you are using git, the .gitignore file tells git to ignore it.
--- a/.npm/plugin/ngAnnotate/npm-shrinkwrap.json
+++ b/.npm/plugin/ngAnnotate/npm-shrinkwrap.json
@@ -1,0 +1,63 @@
+{
+  "dependencies": {
+    "ng-annotate": {
+      "version": "0.15.4",
+      "dependencies": {
+        "acorn": {
+          "version": "0.11.0"
+        },
+        "alter": {
+          "version": "0.2.0"
+        },
+        "convert-source-map": {
+          "version": "0.4.1"
+        },
+        "optimist": {
+          "version": "0.6.1",
+          "dependencies": {
+            "wordwrap": {
+              "version": "0.0.2"
+            },
+            "minimist": {
+              "version": "0.0.10"
+            }
+          }
+        },
+        "ordered-ast-traverse": {
+          "version": "1.1.1",
+          "dependencies": {
+            "ordered-esprima-props": {
+              "version": "1.1.0"
+            }
+          }
+        },
+        "simple-fmt": {
+          "version": "0.1.0"
+        },
+        "simple-is": {
+          "version": "0.2.0"
+        },
+        "source-map": {
+          "version": "0.1.43",
+          "dependencies": {
+            "amdefine": {
+              "version": "0.1.0"
+            }
+          }
+        },
+        "stable": {
+          "version": "0.1.5"
+        },
+        "stringmap": {
+          "version": "0.2.2"
+        },
+        "stringset": {
+          "version": "0.2.1"
+        },
+        "tryor": {
+          "version": "0.1.2"
+        }
+      }
+    }
+  }
+}
--- a/package.js
+++ b/package.js
@@ -15,6 +15,16 @@ Package.registerBuildPlugin({
   }
 });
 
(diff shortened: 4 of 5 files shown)


Urigo:
@netanelgilad :)      I think that's great and it is also optional so why not?
@nickjanssen thank you so much.
Question - do you think about a way to give people the option to use gulp or grunt to run any tasks they want on .ng.js ?
an option to make it more general?

nickjanssen:
@Urigo I'm not sure I understand - I don't think people should be using gulp/grunt at all themselves. Meteor/angular-meteor should care of this in its plugin system, just like it's currently already watching/compiling/minifying files.

As for an option to make it general...perhaps using a config json file? Like `angular-meteor.json`?
Personally I don't think it's necessary, I think the package should be smart enough to figure out on its own what needs to happen. And ng-annotate doesn't do any harm even if you don't need it, so why not let it run always?

Urigo:
@nickjanssen yes that's true, so what if there is a default (with ng-annotate) and then only the programmers that will want to change that (for any reason and I think there might be a few as we can see the great usage and popularity of gulp and grunt) would be able to change that behaviour and use gulp?

nickjanssen:
Again, I don't think people should be using gulp/grunt here. That's Meteor's job. Gulp/Grunt are tools that compile an entire app by concatenation, minification, copying files to a dist folder, watching and serving them, which is exactly what Meteor already does for us. So, we should help Meteor so it is smart enough to handle those `.ng.js` files for us.

As for letting people choose their `annotate` compiler, that's too hardcore (and frankly, absurd. It's not like there are 100 ng-annotate alternatives). Better to choose convention over configuration in this case. `ng-annotate` is the norm and came after `ng-min` which perfectly replicates and improves the behaviour of its predecessor.

Urigo:
@nickjanssen ok, great job!
merging

Urigo:
@nickjanssen  what's your Twitter handle (for the new version announcement)?

nickjanssen:
@Urigo https://twitter.com/nickjanssen_com

pbastowski:
Guys, I'm trying to integrate Babel into my workflow, but running into some trouble. So, what I'm after is:  file.es6 --> babel --> file.js --> ng-annotate --> file.js

The annotate portion is why I'm asking in this thread. Maybe I should open a new thread? 

So, what I did is:

- I added the grigio:babel package to my project - all good so far.
- Next, I created an arrow function and saw it transpiled into a normal ES5 function in the browser. Again, all good.
- Then, I created an angular filter and injected $q into it - this is where things go a bit wrong - ng-annotate has not been ran on the transpiled code.

I have temporarily fixed this by copying grigio:babel and adding an ng-annotate step to it, immediatelu after the babel transpilation, which works very nicely indeed, for the moment. But, I would like to come up with a better solution, if at all possible.

Any suggestions?

netanelgilad:
The problem is with the current implementation of the plugin system that meteor uses. Every file extension can only have one plugin handling it.
That is why the babel package needs your files to end with es6 and not js (cause meteor handles js files) and that is why angular-meteor ng-annotate needs your files to end with .ng.js.
Hopefully this is an issue MDG is going to tackle soon, and we might be able to integrate more than one build plugin per extension.

pbastowski:
@netanelgilad Yes, that's also what I read about the current Meteor plugin design. Do you know if it is possible to define the order in which the plugins execute? I need Babel to run before ng-annotate, because ng-annotate does not understand ES6 as yet.

For the moment, we can work around the one file extension per plugin issue by simply calling several convertors (transpiler, annotation, etc) in sequence within the one Meteor plugin. I was also thinking about creating a generic plugin that would allow me to add several steps to it. Perhaps the plugin itself could have a "plug-in" design, that allows me to hook steps into it. I'll give it more thought...

For now, I will create my own pbastowski:ng-babel package, which processes .es6 files and ng-annotates them all in one. I want to publish it to Atmosphere for others to use, but I have to work out how to publish Meteor packages first.

pbastowski:
My angular-babel package is on Atmosphere. Look for **pbastowski:angular-babel**. My first Meteor package :) Mostly ripped off grigio:babel, though.

netanelgilad:
Awesome @pbastowski.
You should PR the README and add your package to the additional packages.
:beers:

Urigo:
:+1: so good!

jasperkuperus:
For those using mquandalle:harmony and also in need of `ngAnnotate`, check out: [**jasperkuperus:angular-harmony**](https://github.com/jasperkuperus/angular-meteor-harmony).

As my project heavily relies on ES6 `export`, I wasn't able to use `angular-babel` of @pbastowski. Therefore, inspired by @pbastowski, I made a fork of mquandalle:harmony that adds the extra ngAnnotate step.

pbastowski:
Hi @jasperkuperus. Actually, I have a Meteor module `pbastowski:require` that I use to do such things as:

defectModel.es6.js
```javascript
export class defectModel { ... }
```

and in defect.es6.js
```javascript
import defectModel from 'defectModel';
class defect extends defectModel { ... }
```

Would this have helped you?

jasperkuperus:
@pbastowski thanks! I might try that one day! :)

PR #255, opened 2015-03-23
Fix problem with multi-line comments breaking compileAngularTemplates step due to invalid JS generated for the template

smoke:
The issue was introduced in a47ecc2. When a ng.html file contains multi-line comments Meteor failed on the console with error like "Unexpected token ILLEGAL" due to the generated JS that adds the html to the template cache was broken.
For example this could break your app:
```
<div>
<!--
  <div>
      Containing comments on multiple lines breaks the build
  </div>
-->
</div>
```
In order to properly fix that - the best future proof approach is to JSON.stringify the strings as this ensures any JS special character is properly escaped and turned into string literal.

@Urigo - this is real big issue with the latest version v0.8.0, so please act quickly

### DIFF
--- a/plugin/handler.js
+++ b/plugin/handler.js
@@ -15,14 +15,14 @@ Plugin.registerSourceHandler('ng.html', {
     // Ticket here: https://github.com/Urigo/angular-meteor/issues/169
     // A standardized solution to this problem might be on its way, see this ticket:
     // https://github.com/meteor/windows-preview/issues/47
-    '$templateCache.put(\'' + compileStep.inputPath.replace(/\\/g, "/") + '\', \'' +
-      minify(contents.replace(/'/g, "\\'"), {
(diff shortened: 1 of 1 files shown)


Urigo:
@fredrikekelund @netanelgilad I've just reviewed and tested this and looks fine to me.
Please check as well

Urigo:
@smoke amazing! thank you so much - https://github.com/Urigo/angular-meteor/releases/tag/0.8.1

smoke:
@Urigo You are always most welcome!

By the way this code was bugging me from the day I saw it and I was somehow expecting such thing to happen, yet no real issue was present until v0.8.0 was released.

Urigo:
@smoke I wonder what else is bothering you, always feel free to comment!

fredrikekelund:
@smoke thanks for catching that so fast! Sloppily tested on my part

PR #257, opened 2015-03-23
Accounts not Meteor for a few methods

srtucker22:
correct me if i'm wrong, but i believe a bunch of those methods use Accounts, not Meteor. I was getting:

'''undefined is not a function
TypeError: undefined is not a function
    at angularMeteorUser.service.createUser (http://localhost:3000/packages/urigo_angular.js?4b36953210173c874060f2e4feb315d53f8eca68:1562:14)'''


Urigo:
Right!

srtucker22:
thanks for the super fast response!

PR #285, opened 2015-04-11
Updated docs to use $state instead of $location for "AUTH_REQUIRED".

### LINKED ISSUE #283: $meteor.requireUser() doesn't work at all
In [the 8th step of tutorial](http://angularjs.meteor.com/tutorial/step_08) it is said, that I can use
```
resolve: {
    "currentUser": ["$meteor", function($meteor){
      return $meteor.requireUser();
    }]
  }
```
to restrict unauthorized users from any route. But in my application it doesn't work at all. The most interesting part, that in [angular meteor demo application](http://socially-step08.meteor.com/parties) it doesn't work too. I can get to any route, when I'm not logged in.

OK, I've tried to cope with this problem for myself. I've wrote the following code (coffeescript):

```
$stateProvider.state('signIn',
      url: '/sign-in'
      controller: 'signInCtrl'
      templateUrl: 'client/pages/users/views/signIn.html'
    ).state('home',
      url: '/'
      controller: 'indexCtrl'
      templateUrl: 'client/pages/index/views/index.html'
      resolve:
        currentUserExists: currentUserExists
    )

currentUserExists = ['$q', '$timeout', '$state', '$rootScope', '$log', ($q, $timeout, $state, $rootScope, $log) ->

      if $rootScope.currentUser
    # Resolve the promise successfully
        $q.when()
      else
    # The next bit of code is asynchronously tricky.
        $timeout ->
    # This code runs after the authentication promise has been rejected.
    # Go to the log-in page
          $state.go 'signIn'
          return
        # Reject the authentication promise to prevent the state from loading
        $q.reject()
    
]

```

It checks, if currentUser exists. If not, redirect to singIn page.
The problem is, that on site initial load (when I refresh a page or when I type address in address bar), ``` $rootScope.curentUser ``` returns ``` undefined ``` (not  ``` null ``` ), but when I navigate through site, everything is OK. So I've edited code to wait for one second, if is the first load. But this code is really ugly:
```

currentUserExists = ['$q', '$timeout', '$state', '$rootScope', '$log', ($q, $timeout, $state, $rootScope, $log) ->
  if window.notFirstTime == true
    $q.when()
  else
    $timeout ->
      window.notFirstTime = true
      if $rootScope.currentUser
    # Resolve the promise successfully
        $q.when()
      else
    # The next bit of code is asynchronously tricky.
        $timeout ->
    # This code runs after the authentication promise has been rejected.
    # Go to the log-in page
          $state.go 'signIn'
          return
        # Reject the authentication promise to prevent the state from loading
        $q.reject()
    , 1000
]

```

So the questions are:
* Why ``` $meteor.requireUser() ``` doesn't work as excepted?
* Why ``` $rootScope.currentUser ``` returns undefined on site load?

goooseman:
Closes #283

### DIFF
--- a/.docs/angular-meteor/client/views/api/api.auth.html
+++ b/.docs/angular-meteor/client/views/api/api.auth.html
@@ -148,12 +148,13 @@ See the <a href="/tutorial/step_08#authenticationwithrouters">“Authentication 
         }
     });
 
-    app.run(["$rootScope", "$location", function($rootScope, $location) {
+    app.run(["$rootScope", "$state", function($rootScope, $state) {
       $rootScope.$on("$stateChangeError", function(event, toState, toParams, fromState, fromParams, error) {
         // We can catch the error thrown when the $meteor.requireUser() promise is rejected
         // and redirect the user back to the login page
         if (error === "AUTH_REQUIRED") {
-          $location.path("/login");
+          // It is better to use $state instead of $location. See Issue #283.
+          $state.go('logIn')
         }
       });
     }]);


Urigo:
Thanks!

Urigo:
@goooseman can you please do the same fix to the tutorial in step 8?

goooseman:
@Urigo Not a problem

PR #287, opened 2015-04-11
Update tests to work with the latest Angular version

panw:
ng-if  does not work properly on img tags in 1.3.13 and updating to angular 1.3.15 has the fix. See issue: https://github.com/angular/angular.js/issues/7183


Urigo:
@panw angular-meteor already uses the latest Angular:
https://github.com/Urigo/angular-meteor/blob/master/package.js#L32

but thanks for updating the tests!

PR #288, opened 2015-04-12
Collection rendering speed improvement

barbatus:
Collection rendering speed improvement based on an async apply to combine multiple operations made to the collection.


Urigo:
@barbatus this is very nice.
The problem is what will happen if during that time there will be updates from the client?
in this case, the $watch might miss those updates and it will cause a mess..

barbatus:
@Urigo Sorry, which $watch do you mean?
That PR doesn't change collection itself it just adds accumulative $apply which controls visual aspect. So data consistency should persist if any $watch or whatever happens.

Urigo:
@barbatus this watch:
https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-meteorCollection.js#L237-L277

it does matter because of the logic of the watch (with UPDATING_FROM_SERVER flag)

barbatus:
@Urigo Ok (not diving into logic of a potential issue). This [line](https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-meteorCollection.js#L139) stays intact in the current change, while ````self.UPDATING_FROM_SERVER = false;```` is set in ````setTimeout```` same as in the current code where it's set in  ````$timeout```` which is same 10-15ms. So question is why you are sure that something (you are talking about) won't happen in this 10-15ms and will happen in interval approx. twice large (I believe in average collection after-subscribe fill-out takes approx. time, say, 20ms) 30-35ms?

netanelgilad:
@barbatus I think you are right about the $timeout in the current code. I think it should not be there any way and it should be immediately after the apply. But the bug that could happen is still the same.
We thought about it, and when we aggregate server updates we can never be sure that we are not also collecting updates from the client and thinking they came from the server.
That is an interesting issue and we are still looking for a solution. It might come to a place where we have to decide if we prefer having better performance of server updates over the tiny chance we might miss a client update.

Urigo:
@barbatus like @netanelgilad said, the problem is not with catching the server updates but catching client updates and understanding they happen.
If between intervals there will be an update from the client, we won't know it happen from the client and not from the server.
In this case I prefer concurrency over performance.  
But, your change can be great when the user sends the false flag to autobind.
In this case we don't care about watching the client side automaticly and we can do that interval with no problem.
Do you think you want to try adding that only when the false flag is false?

barbatus:
@urigo @netanelgilad That makes sense too.
I also tried to solve it somehow by backing server updates into another one auxiliary collection.
Please, take a look at the last commit and demo deployed [http://ng-meteor.meteor.com] (http://ng-meteor.meteor.com). I tried to emulate some concurrency as you can see [here] (https://github.com/barbatus/ng-meteor/blob/master/client/views/pages/main_page.js); it just pushes new elements at random intervals before and after updates from server trying to hit an interval when updates are happening. It seems to work correct now. So if you take a look at the console, this line ````diffArray item saved foo added```` appears exactly 10 times as expected no matter when those setTimeouts triggered.
There also seems to be a bug [here](https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-meteorCollection.js#L256):  ````ngCollection.pop()```` is needed instead.
As you can see also a duplication of ````$rootScope.$apply()```` is needed [here](https://github.com/barbatus/angular-meteor/commit/502ccd29c69a312f75620db28f0b12bf86c646a6#diff-32cf712277b42c3ac88358aa2911a7f4R169) to work properly  but it can be solved by moving out  ````diffArray```` method in that collection watch into a separate method and then we can save all the differences before ````$rootScope.$apply()````.

barbatus:
@Urigo So if you'd like to skip what's been done with another one collection, then we can just stop on changing this PR to only when 'autobind is set false', no big deal.

Urigo:
@barbatus really impressive work!

1. I love to _serverBackup change instead of old items, looks nicer but is it faster than copying it once like before?
2. The reason for the slice was to revert the client side change so we will get it from the server side (in case of pagination, sorting and other cursor queries). can you explain better the difference of your implementation?  in case there are many additions?
3. About the concurrency checking, I would want to have maybe another client or the server itself updating the same data in the same time and not just one client. 
When I ran 2 different clients on your website, it looks like that:
screen shot 2015-04-16 at 6 56 27 pm

but I guess it's because of the limit...

also do you think we can add a test that does that?

Urigo:
@barbatus if it can work with the autobind I do want to continue trying

Urigo:
@barbatus In my last message with the feedback and the photo I meant let's continue working on that, not to put it aside!

barbatus:
1. This collection is required if you want to know what new items have been inserted.
    Collection itself can be modified and this._serverBackup always preserves items that come from the 
    server so we can compare and find out what's been added from client.
2. Yes, I understand the reason. But when you compare two arrays and get a list of indices, they always stay the same, at the same time ````splice``` modifies collection so every next index needs to be updated.

Urigo:
@barbatus cool thank you so much for the explanations.
I hope you understand why I'm being very careful about this change..

Do you think we can add automatic tests for that?

barbatus:
@Urigo I am almost done with tests and little new changes

Urigo:
@barbatus I'm really looking forward for this!
I will also love to chat with you about that in video if you want, you can find my email on my profile

barbatus:
@Urigo Please take a look at the commit ecc500f before the last merge (my bad, I bit messed up with commits).
You can find there an integration test (server involved) that verifies concurrent changes.
It seems can be done partly with mocks but I didn't how to mock collection's observer so I added real publish. It works as now, if I find a simpler approach I'll change after.

Urigo:
@barbatus ok let's do this!
I've reviewed it now and there are so many good stuff here.
As there are a lot, I want to have a short meeting going through those with you so we won't need to split the pull request.
When can you meet???

bower.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "angular-meteor",
   "main": "angular-meteor.js",
-  "version": "0.8.6",
+  "version": "0.8.7",
```
Urigo: Just leave the version numbers out so we won't need to change those everytime

modules/angular-meteor-meteorCollection.js
```diff
@@ -28,15 +28,6 @@ angularMeteorCollections.factory('AngularMeteorCollection', ['$q', '$meteorSubsc
       function upsertObject(item, $q) {
         var deferred = $q.defer();
```
Urigo: Why did you remove it?  
Can you add it so that this will be in?
https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-meteorCollection.js#L31-L38

modules/angular-meteor-meteorCollection.js
```diff
@@ -142,15 +132,28 @@ angularMeteorCollections.factory('AngularMeteorCollection', ['$q', '$meteorSubsc
...
-        // Making sure we are setting to false only after one digest cycle and not before
-        $timeout(function(){
+        // while handling updates from the server.
+        if (!self.UPDATING_FROM_SERVER) {
+          self.UPDATING_FROM_SERVER = true;
+          $rootScope.$apply();
+        }
+        promise = $timeout(function () {
+          // Saves changes only when new docs have been added.
+          if (self.length != self._serverBackup.length) {
```
Urigo: What about update specific object?

modules/angular-meteor-meteorCollection.js
```diff
@@ -239,48 +248,25 @@ angularMeteorCollections.factory('$meteorCollection', ['AngularMeteorCollection'
...
           ngCollection.unregisterAutoBind = $rootScope.$watch(function () {
-            if (ngCollection.UPDATING_FROM_SERVER){
-              realOldItems = _.without(ngCollection, 'UPDATING_FROM_SERVER');
+            if (ngCollection.UPDATING_FROM_SERVER) {
               return 'UPDATING_FROM_SERVER';
             }
             return _.without(ngCollection, 'UPDATING_FROM_SERVER');
           }, function (newItems, oldItems) {
-            if (newItems == 'UPDATING_FROM_SERVER')
+            if (newItems == 'UPDATING_FROM_SERVER' ||
```
Urigo: But what happens if the oldItems is 'UPDATING_FROM_SERVER' and the new items are new collection?
it means we got a change from the client and here you are immediately return without updating the server...

bower.json
barbatus: Fixed by update to the head.

modules/angular-meteor-meteorCollection.js
barbatus: Fixed by update to the head.

modules/angular-meteor-meteorCollection.js
barbatus: I removed the line ````self.length != self._serverBackup.length````
Now all changes happened (incl. just updates) to the objects on the client should be handled in the updateCollection in the $timeout.

modules/angular-meteor-meteorCollection.js
barbatus: oldItems is 'UPDATING_FROM_SERVER' and new items is a collection is exactly when we set 
self.UPDATING_FROM_SERVER to false  and then apply root scope (see comment above). That means collection should have been saved already  right before in that updateCollection function ([here](https://github.com/Urigo/angular-meteor/commit/7b549c2db91763eaa770282ca20349063833336b#diff-32cf712277b42c3ac88358aa2911a7f4R160))

modules/angular-meteor-meteorCollection.js
```diff
@@ -239,48 +255,25 @@ angularMeteorCollections.factory('$meteorCollection', ['AngularMeteorCollection'
...
               return;
 
-            if (oldItems == 'UPDATING_FROM_SERVER')
-              oldItems = realOldItems;
-
-
             if (newItems !== oldItems) {
+              ngCollection.unregisterAutoBind();
+
+              updateCollection(ngCollection, oldItems, diffArray);
```
Urigo: That does save code but it also stopping and setting the watch even when it is not necessary, like on update and delete by the client

barbatus: This part seems to have not been changed if you mean ````ngCollection.unregisterAutoBind````, I only wrapped code into updateCollection. I'd fix it in a different PR.

Urigo: Yes I've meant the ngCollection.unregisterAutoBind, it's better to do it only for the addedAt..
Can't you change it now?

Urigo:
Thank you so so much @barbatus

PR #293, opened 2015-04-13
Bugfix/cfs error

epotvin:


Urigo:
@epotvin thank you so much!
why did you change the internalProps and how does it affect performance of non CollectionFS objects?

epotvin:
The idea was to remove functions from the object properties when the angular digest process was running to avoid the same problems the internalProps were removed for. This seems to give it a really hard time. It may not be the best approach but it solves the problem.

About performance, I didn't test it thoroughly, but I suppose it's not the best approach for that either. The list of the omitted fields could be cached in some map with prototype name used as key or something like that, but functions can be added afterward so it would not work anyway...

I think it's important to avoid coupling the used class (in this case FS.File) with the code, but we should also be able to omit properties in deeper level if needed. It's not the case here but suppose we have a field named 'whatever' which is an object containing other fields we want to be part of the change comparison, but not every one of them. We would like to remove whatever.subproperty...

I'm thinking out loud anyway... let me do some tries and come back with more details.

Urigo:
that's an interesting idea, I need to look deeper into that, but in any way I think it's good to separate the changes.
Let's only handle the CollectionFS here and handle the properties in a separate PR

Urigo:
@epotvin what are you thoughts about separating the two?

epotvin:
Hi Uri! Sorry for the delay, I didn't have a lot of time lately, but I'll try to fix that today. Actually, I think that without the change for the internalProps the fix for the collection doesn't work at all. But let me make some tests and come back with more details soon. Of course it would be easier to follow changes if they were separated, so maybe I could fix the internalProps issue first.

epotvin:
So... I'm not sure what happened but I'm unable to reproduce the problem that made me code that change in the first place... I still have to change to change to setAutoBind function in the collection class to keep angular from copying the collection property of the File object, but the logic removing functions from object happens to be useless... I fear the problem will eventually jump back in my face but everything works well with the actual code.

I added some failures in the tests thought. I'll look at this tonight. Right now I have to leave.

Manu

Urigo:
Cool, can't wait and thank you so much

Urigo:
Thanks!

PR #296, opened 2015-04-14
add pbastowski:angular-babel

pbastowski:
Added a reference in README to the new package pbastowski:angular-babel. JavaScript files ending in .es6 will first be transpiled by Babel and then annotated with ng-annotate. Without this package angular-meteor will only ng-annotate files that end in .ng.js.


Urigo:
good job!

PR #301, opened 2015-04-16
Allow to use a checker function before resolving

auterium:
Current behaviour: when using $meteor.requireUser(), a resolved promise will be returned if there is an authenticated user, and a rejection if not.

Proposed behaviour: allow $meteor.requireUser() to receive a function that will evaluate if the logged user has permission or not.

Usage:
The received function will be executed with the user object as the only paramenter. If the function returns a string, such string will be used to reject the resolution; otherwise it will resolve with the user as expected.

Example:
```javascript
$stateProvider.state('secure-page', {
  url: '/my-top-secret-page',
  templateUrl: 'my/top/secret/page.ng.html',
  resolve: {
    currentUser: ['$meteor', function($meteor) {
      return $meteor.requireUser(function(user) {
        if(!_.contains(user.roles, 'root')) {
          return 'FORBIDDEN';
        }
      });
    }]
  }
});
```

Use cases:
- Role based access control
- Check for active/suspended account

Regards,
Marc

### DIFF
--- a/.docs/angular-meteor/client/views/api/api.auth.html
+++ b/.docs/angular-meteor/client/views/api/api.auth.html
@@ -22,7 +22,7 @@ Methods and [$rootScope](https://docs.angularjs.org/api/ng/service/$rootScope) v
 
     $meteor.waitForUser()
 
-    $meteor.requireUser()
+    $meteor.requireUser([validatorFn])
 
     $meteor.loginWithPassword(user, password)
 
@@ -119,7 +119,7 @@ See the <a href="/tutorial/step_08#authenticationwithrouters">“Authentication 
     });
 
 <br>
-<h3><p><code><span class="pln">requireUser();</span></code></p></h3>
+<h3><p><code><span class="pln">requireUser([validatorFn]);</span></code></p></h3>
 
 Resolves the promise successfully if a user is authenticated and rejects otherwise.
 <br>
@@ -128,6 +128,10 @@ This is useful in cases where you want to require a route to have an authenticat
 You can catch the rejected promise and redirect the unauthenticated user to a different page, such as the login page.
 <br>
 See the <a href="/tutorial/step_08#authenticationwithrouters">“Authentication with Routers”</a> section of our tutorial for more information and a full example.
+<br>
+This method can receive a function as a parameter, which can be used to do custom validation prior to resolving/rejecting the promise.
+<br>
+If the validator function is present, the method will only resolve if the function returns true. If it returns a string, it will reject the promise with such string. Otherwise (false, null, undefined) it will reject the promise with the "FORBIDDEN" error string.
 
 
 
@@ -146,16 +150,53 @@ See the <a href="/tutorial/step_08#authenticationwithrouters">“Authentication 
             return $meteor.requireUser();
           }]
         }
-    });
+      })
+      .state('admin', {
+        url: '/admin',
+        templateUrl: 'client/views/admin.ng.html',
+        controller: 'AdminController'
+        resolve: {
+          "currentUser": ["$meteor", function($meteor){
+            return $meteor.requireUser(function(user) {
+              return user.username==='admin';
+            });
+          }]
+        }
+      })
+      .state('admin2', {
+        url: '/admin2',
+        templateUrl: 'client/views/admin2.ng.html',
+        controller: 'Admin2Controller'
+        resolve: {
+          "currentUser": ["$meteor", function($meteor){
+            return $meteor.requireUser(function(user) {
+              if (user.username==='admin2') {
+                return true;
+              }
+
+              return 'UNAUTHORIZED';
+            });
+          }]
+        }
+      });
 
     app.run(["$rootScope", "$state", function($rootScope, $state) {
       $rootScope.$on("$stateChangeError", function(event, toState, toParams, fromState, fromParams, error) {
         // We can catch the error thrown when the $meteor.requireUser() promise is rejected
-        // and redirect the user back to the login page
-        if (error === "AUTH_REQUIRED") {
-          // It is better to use $state instead of $location. See Issue #283.
-          $state.go('logIn')
-        }
+        // or the custom error, and redirect the user back to the login page
+        switch(error) {
+          case "AUTH_REQUIRED":
+            $state.go('logIn');
+            break;
+          case "FORBIDDEN":
+            $state.go('forbidden');
+            break;
+          case "UNAUTHORIZED":
+            $state.go('unauthorized');
+            break;
+          default:
+            $state.go('internal-client-error');
+		}
       });
     }]);
 
--- a/.docs/angular-meteor/client/views/api/api.auth.html
+++ b/.docs/angular-meteor/client/views/api/api.auth.html
@@ -22,7 +22,9 @@ Methods and [$rootScope](https://docs.angularjs.org/api/ng/service/$rootScope) v
 
     $meteor.waitForUser()
 
-    $meteor.requireUser([validatorFn])
+    $meteor.requireUser()
+
+    $meteor.requireValidUser(validatorFn)
 
     $meteor.loginWithPassword(user, password)
 
@@ -119,7 +121,7 @@ See the <a href="/tutorial/step_08#authenticationwithrouters">“Authentication 
     });
 
 <br>
-<h3><p><code><span class="pln">requireUser([validatorFn]);</span></code></p></h3>
+<h3><p><code><span class="pln">requireUser();</span></code></p></h3>
 
 Resolves the promise successfully if a user is authenticated and rejects otherwise.
 <br>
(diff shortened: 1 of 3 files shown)


Urigo:
@auterium amazing work and a great idea!

1. Do you think returning a string when it's false in intuitive?
I was thinking maybe doing it more similar to the [allow-deny](http://docs.meteor.com/#/basic/Mongo-Collection-allow).  like:
- false is will reject with a default string
- true will return the user 
- string will reject do what you have done now
- undefined or nothing reject with a default string
I think it will make it a little more robust, what do you think?

2. Can you add documentation to the tutorial and API?
https://github.com/Urigo/angular-meteor/edit/master/.docs/angular-meteor/client/views/api/api.auth.html
https://github.com/Urigo/angular-meteor/edit/master/.docs/angular-meteor/client/views/steps/tutorial.step_08.html

Thank you so much

auterium:
@Urigo thanks for the quick feedback!

I used this approach as is the one used in [meteor-simple-schema's custom validation] (https://github.com/aldeed/meteor-simple-schema#custom-validation).

I agree that it sounds more intuitive to return a true/false/string-reason, but I think it helps reduce the amount of LoC to use it the way I did. It's not much, about 3-5 lines per each checker function, but that could have an impact in a big project.

I think that the final answer on how to approach this would be: how is it going to be used around the whole project? If the true/false/string-reason pattern is used through the project, then this should be the one used and the code modified.

Once we agree on the implementation, I'll be happy to update the docs.

Regards,
Marc

Urigo:
Sorry for the late response..

Hmmmm, so the difference is actually what will undefined and empty will result in (because adding true as an acceptable result shouldn't add more code I think), like, what question does the function answers:

1. Who can go in here?
2. Who cannot go in here?

I think usually the security approach is to define who can and not who can't, also maybe the more secure and error proof (security is a different case than scheme checking I think).

About the whole project approach, it's a good idea but I don't think we have any at the time :)

Maybe you can write scenarios and let's compare, then we can get a better look at what's the real difference (anyway let's not hang too much with it, it's a great change, I really want it in and I won't bother you with it)

auterium:
No worries on the late response. We all have paid work that's bigger priority than contributing to opensource projects (unless you're a billionaire that has nothing better to do than contributing, but I don't think we're that).

One of the reasons why I went with this approach was to just return the error cause and didn't wanted to impose a default one. But giving it a second thought, the default one would be "FORBIDDEN" when returning false, undefined or null, as you already defined the "AUTH_REQUIRED" for when no user logged in. So I'll guess I'll change it to that approach (not a biggie).

I'll work on it through the day along with the documentation and update the PR maybe tonight.

Regards,
Marc

Urigo:
@auterium ohh you are right, haven't thought about that actually.
very cool.
Thank you so so much

smoke:
I am not completely into the use case, but I think there is no need to complicate the $meteor.requireUser function.
Since it is returning a promise - you can chain the promise and have the function run on the chain. For example:
```
$stateProvider.state('secure-page', {
  url: '/my-top-secret-page',
  templateUrl: 'my/top/secret/page.ng.html',
  resolve: {
    currentUserAccessGranted: ['$meteor', function($meteor) {
      return $meteor.requireUser().then(function(user) {
        if(!_.contains(user.roles, 'root')) {
          return 'FORBIDDEN';
        }
      });
    }]
  }
});
```

Using promises have the following advantages:
* Does not complicate the return types of the $meteor.requireUser()
* Does not complicate the code of $meteor.requireUser() itself
* Allows you to chain other promises - for example you can even check the requirements against external service using $http request - see https://docs.angularjs.org/api/ng/service/$q#chaining-promises

auterium:
@smoke what you propose will allways resolve, either an undefined or string value, but always access the controller.

The proposed code will not change the fact that $meteor.requireUser() will return a promise and that you can call it without arguments. What changes is that you can add extra validation to be run before returning a resolved promise and that can also reject such promise.

The idea behind this is to be able to reject the promise so that the controller/route the user is trying to access will never be reached. Also, by doing a promise rejection, this will cause a $stateChangeError for which there's already a listener, it will only need another if statement. Here's the example for that:

```javascript
angular.module('my-module')
.run(['$rootScope', '$state', function($rootScope, $state) {
	$rootScope.$on('$stateChangeError', function(event, toState, toParams, fromState, fromParams, error) {
		if(error==='AUTH_REQUIRED') {
			$state.go('login');
		}

		// This is the handler for a forbidden rejection
		if(error==='FORBIDDEN') {
			$state.go('forbidden');
		}
	});
}]);
```

As you can see in the code above, it's the same as the one you find in the tutorial, so you already have that code, but just add the other error handler for the FORBIDDEN rejection error and that's all the code you need to handle all unauthorized accesses.

With your proposal you'll have to do the check for the "currentUserIsAccessGranted" on each controller and from there do the redirection. This is a bad practice because:
- It would make you repeat your code over each controller
- It shouldn't be the controller's responsability to determine if the user has permission to access it or not

Regards,
Marc

smoke:
@auterium Sorry that I didn't explained it the first time - with the chained promises you can return rejected promise, which would do the job.

A bit better example:
```
$stateProvider.state('secure-page', {
  url: '/my-top-secret-page',
  templateUrl: 'my/top/secret/page.ng.html',
  resolve: {
    currentUser: ['$meteor', '$q', function($meteor, $q) {
      return $meteor.requireUser().then(function(user) {
        if(!_.contains(user.roles, 'root')) {
          // fail the promise chain
          return $q.reject('FORBIDDEN');
        }
        // keep the success promise chain
        return user;
      });
    }]
  }
});
```

As for the points you have listed - I totally agree, just didn't read the complete use case and pull request, to reveal the mechanics.

auterium:
@smoke thanks for the example, I wasn't aware that could be done (though it totally makes sense).

@Urigo I've updated the pull request with the accorded usage plus the API docs. I didn't updated the tutorial as it requires multiple changes to it to add extra pages/processes that would make sense to add an extra validation to the state resolving of the user.

netanelgilad:
I actually like @smoke solution better. To me, it feels more angular-oriented and a better separation of concerns. If the developer wants to do more checks, chaining promises makes total sense...
As I see it, all of the use cases listed for the validator function can be achieved with chaining promises. Am I missing something?

auterium:
@netanelgilad I agree it feels more angular-oriented, and this is where a lot of discussions will come over as the angular-way and the meteor-way are different and we're merging them together. This code is on the angular part of the project and I agree it should be used the angular way.

I have 2 reasons to use a validator function over a promise chaining:
- No need for extra injection: Within the $meteor.requireUser() method there's already an instance of $q injected, so not using it and instead doing another inject elsewhere feels wrong for me.
- Code simplicity: It requires a bit less lines of code and I think it's a bit easier to understand for people that are just getting started using angular and/or meteor.

Also, this code doesn't require you to use it this way. It still returns a promise and can be called without any args, so you can do promise chaining with or without a validation function

netanelgilad:
@auterium You make good points but I am still not convinced :).
I prefer that $meteor.requireUser will have a single responsibility and to use the mechanism that already exists to allow to augment it. The less we impose on the developer, the better (like if undefined is returned or the other options mentioned above).
And it is one less source for bugs ;)
Any way, this is a good discussion!
@Urigo and any one else, your thoughts here welcome.

auterium:
@netanelgilad the proposal doesn't try to impose anything on de developer as it is not mandatory to pass such argument, it's just a helper. If the developer wants to use promise chaining instead of using the validation function, there's nothing that stops them.

I'm not such a puritan on that every method must have a single responsability, but if this is the case, then the option would be to create another method called $meteor.requireValidUser() which will require such function:
```javascript
angularMeteorUser.service('$meteorUser', ['$rootScope', '$meteorUtils', '$q',
  function($rootScope, $meteorUtils, $q){
    var self = this;

    this.requireUser = function(){
      // ommited for comment
    };

    this.requireValidUser = function(validatorFn){
      return self.requireUser().then(function(user){
        var valid = validatorFn( user );

        if ( valid === true )
          return user;
        else if ( typeof valid === "string" )
          return $.reject( valid );
        else
          return $q.reject( "FORBIDDEN" );
      });
  };

  // Rest of the service's methods
}]);

Urigo:
Guys, this is a great discussion!

@auterium I think your last suggestion answers everyones needs.  can you please change the pull request to that and I will merge it?

auterium:
@Urigo I've updated the code and the docs

Urigo:
Nice job, I will also add this to the tutorial soon

smoke:
@auterium I believe there is a small typo you need to change ```$.reject``` -> ```$q.reject```

auterium:
Thanks @smoke I've created the new pull request for it

Urigo:
@auterium can you update the examples?
I think they are still calling requireUser and not requireValidUser

PR #303, opened 2015-04-17
Fix #258 - Add compileStep.packageName to templateCache key

### LINKED ISSUE #258: Add compileStep.packageName to templateCache key
I like to organize my angular templates into meteor packages to make it easier to use tinytest, etc.

The angular-meteor plugin uses compileStep.inputPath when placing the compiled templates into $templateCache.  This causes templates with the same path (relative to their package) to register with the same key in $templateCache.

Any chance of fixing this? (it would be a breaking change)

pward123:
***BREAKING CHANGE*** templateUrl paths for *.ng.html files inside meteor packages have been changed.

angular-meteor now prefixes the package name when adding templates to $templateCache. The colon in the package name is replaced with an underscore and the package name is separated from the template name with an underscore. So, if you have a package named my-app:my-package with a template at packages/my-app:my-template/client/views/my-template.ng.html, the templateUrl you use in the route should be:

    my-app_my-package_client/views/my-template.ng.html

### DIFF
--- a/plugin/handler.js
+++ b/plugin/handler.js
@@ -12,13 +12,17 @@ Plugin.registerSourceHandler('ng.html', {
   // Just parse the html to make sure it is correct before minifying
   HTMLTools.parseFragment(contents);
 
+  // Build the templateCache prefix using the package name
+  var packagePrefix = compileStep.packageName;
+  packagePrefix = packagePrefix ? (packagePrefix.replace(/:/g, '_') + '_') : '';
+
   var results = 'angular.module(\'angular-meteor\').run([\'$templateCache\', function($templateCache) {' +
     // Since compileStep.inputPath uses backslashes on Windows, we need replace them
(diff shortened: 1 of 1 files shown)


Urigo:
Thank you so much!
Where do you think we can specify this in the API docs?
Can you add it to the PR?
also, do you think maybe somewhere in the tutorial might be a good idea?

pward123:
Obviously, it should end up in the change log. I don't think the main repo README.md needs to change. Instead, I'd probably add a bullet point or paragraph describing templateUrl in the **Tutorial - Routes definition** section. I can take a stab at this if you like.

Urigo:
Yes that will be great please try.
expand as much as you like on the build process

Urigo:
Thank you so much!

I've added a bit of documentation [here](https://github.com/Urigo/angular-meteor/commit/092c522931d57744395d5e41223a3c559615708a), let me know if that works...

PR #307, opened 2015-04-21
Reverts back recent changes to the onAfterAction handler.

barbatus:
After having tested my app, I decided that recent changes to the onAfterAction are too complex.
I propose to revert it back with once change: check it's been compiled only once. It's what regular Angular needs. Anything else is up to a developer. For example, one could create a separate Angular module with onAfterAction (or whatever) and write there any route changes handling (as I did).


Urigo:
@barbatus you are the iron:route boss :)

PR #313, opened 2015-04-28
Update docs to use requireValidUser()

auterium:
As per @Urigo request


Urigo:
@auterium thanks!
maybe you should also document that there is no need to actually send the 'user' parameter?
also, maybe just remove it for the example? will it still work?

auterium:
@Urigo I can't think of a scenario where to use it without the user argument

Urigo:
@auterium but the user argument is supplied inside anyway no?

auterium:
Ohh you mean there's no need to do the
```javascript
return $meteor.requireValidUser()
```
stuff?

The callback that the function receives will always be executed pasing the user as the only argument

Urigo:
yes.. so like add a comment about it or something so people will understand where is it coming from

PR #320, opened 2015-05-01
object check in $$hashKey deletion in nested values

mxab:
Hi, I had and object that contained a null value like this:
```
myCollection.save({foo: null})
```
which produced an error at ```delete prop.$$hashKey```


Urigo:
Thanks @mxab

PR #321, opened 2015-05-02
Remove Blaze view handler on $scope destroy.

barbatus:


Urigo:
@barbatus you are the Blaze boss  :)
Can you explain so I can learn?

barbatus:
@Urigo I think it's pretty clear here. Template is created and rendered dynamically in this directive so we need to remove Blaze reference to it at some point (e.g. when new page is opened) otherwise it will dangle. $destroy is standard Angular event emitted when scope destroyed.

Urigo:
oh for some reason I so it the other way around.
Yes totally obvious.
Thanks!

Issue #304, opened 2015-04-17
Better handle opening and closing subscriptions


Urigo:
We want to help developers handle subscriptions more easily.

We have internal discussions about that and also many people from the community started coming up with interesting solutions.

I'm opening this issue so that we will move the discussion here so everyone can affect the solution.

Starting point:

Right now there are 2 general approaches with different solutions for each of them:
1. Router level
2. Controller level

I personally tend to go with the controller level as it's a more general solution (closer to component level subscription like Blaze and Angular 2.0) but it has it's problems and I'm not sure what the actual technical solution even be.
In routers I've thought and also heard many solutions so it might be easier to implement.

Looking forward for your thoughts

revov:
@Urigo, what works for me is a simple Factory that I use in my project to help me with the huge overhead of cleaning up subscriptions, collections and objects:

```javascript
angular.module('myModule')
    .factory('meteorWrapperFactory', [
        '$meteor', '$q',
        function($meteor, $q) {
            return {
                subscribe: function( scope, subscriptionParams /*variadic*/ ) {
                    var deferred = $q.defer();
                    var args = _.takeRight(arguments, arguments.length-1);
                    args.push({
                        onReady: function () {
                            deferred.resolve(subscriptionHandle);
                        },
                        onError: deferred.reject
                    });

                    var subscriptionHandle = Meteor.subscribe.apply( this, args );
                    scope.$on('$destroy', function() {
                        subscriptionHandle.stop();
                    });

                    return deferred.promise;
                },
                collection: function( scope, collectionParams /*variadic*/ ) {
                    var collection = $meteor.collection.apply( this, _.takeRight(arguments, arguments.length-1) );
                    scope.$on('$destroy', function() {
                        collection.stop();
                    });

                    return collection;
                },
                object: function( scope, objectParams /*variadic*/ ) {
                    var object = $meteor.object.apply( this, _.takeRight(arguments, arguments.length-1) );
                    scope.$on('$destroy', function() {
                        object.stop();
                    });

                    return object;
                }
            };
        }
    ]);
```

Note that I am using lodash to get the variadic array of arguments, but this can easily be achieved with vanilla Javascript (considering Meteor's deprecation of Underscore js). This is a really simple solution that won't fit all needs, but it gets rid of the awful cleanup overhead while still preserving performance and providing flexibility and ease of use. What I like about this approach is that it doesn't enforce architectural decisions on the developer. It just encapsulates the tedious boilerplate code.

[...]

PR #323, opened 2015-05-03
$scope subscribe, collection, and object methods prefixed with $meteor

auterium:
As discussed in #304

### DIFF
--- a/.docs/angular-meteor/client/views/api/api.meteorCollection.html
+++ b/.docs/angular-meteor/client/views/api/api.meteorCollection.html
@@ -8,15 +8,19 @@
     <do-nothing>
         {{#markdown}}
 
-# $meteor.collection
+# $meteor.collection   /    $scope.$meteorCollection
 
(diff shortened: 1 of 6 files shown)


auterium:
@Urigo I know you're packed with work, but any idea if this is going through?

Urigo:
Hi sorry!
I'm in SF meeting with Meteor and Angular, a lot of really good things are coming.
Tonight I will go through your pull request!

auterium:
That's cool. Thanks for the reply. I wanted to know as I'm already using this things for projects, and I would like to use it directly from the published package and not from a local one.

Urigo:
That great, it make me more relaxed on merging it
Good job!

auterium:
Thanks! Are there any upcoming meetings you're going to in the US? When I heard about ng-vegas it was too late for me to buy a ticket or anything, but it would be nice to go to one and meet.

Urigo:
Unfortunately I'm flying out this Saturday..
But I'll probably be back soon

PR #324, opened 2015-05-04
Update tutorial.step_18.html

MilosStanic:
Angular-material redesign of socially app


Urigo:
Amazing!
will go through it soon and deploy

PR #326, opened 2015-05-05
(site): move navbar items

netanelgilad:
@Urigo check this out. Saw this on another site and thought it could be better than a dropdown which is less likely to be clicked.


Urigo:
@netanelgilad yes it's better!

PR #328, opened 2015-05-06
Add a basic page explaining Angular 2 intigration

simonv3:
# What's Changed
* Added a simple page explaining Angular 2 Integration options.
* Removed bolding from header because it had a weird effect in Firefox.

screenshot 2015-05-06 12 37 11

# What's Next
* It might be a good idea to simplify the example code for each section, keeping it bare bones.


simonv3:
Hmm, I'm not sure why this would be failing. Didn't change anything :/

PR #330, opened 2015-05-07
Update of Step 18 material design login

MilosStanic:
Second attempt to see if we can automerge


Tallyb:
@MilosStanic  wouldn't it be more generic if you change the path on logout to "/" rather than to state 'parties'? then the code can be copied directly (basis for the ui-accounts).

MilosStanic:
@Tallyb I don't see why not.. I guess the app will auto redirect to  `/parties` when redirected to `/`. But anyway, this is a special case. If I go into `ui-accounts` I will do it properly.

PR #331, opened 2015-05-07
(site): return navbar changes

netanelgilad:


Urigo:
@netanelgilad really????

PR #338, opened 2015-05-12
trivial word addition complete -> complete support

ppotoplyak:
Hi

I think there is a word missing in the User Auth Readme section.  Trivial change.

thanks


Urigo:
Thanks for the catch!

PR #342, opened 2015-05-16
added $meteorSubscribe to AngularMeteorObject and $meteor service

alansikora:
Added $meteorSubscribe to AngularMeteorObject and $meteor service, so it is possible to use the same method with $meteor and $scope

### DIFF
--- a/angular-meteor.js
+++ b/angular-meteor.js
@@ -36,6 +36,7 @@ angularMeteor.service('$meteor', ['$meteorCollection', '$meteorObject', '$meteor
     this.collection = $meteorCollection;
     this.object = $meteorObject;
     this.subscribe = $meteorSubscribe.subscribe;
+    this.$meteorSubscribe = $meteorSubscribe.subscribe;
     this.call = $meteorMethods.call;
     this.loginWithPassword = $meteorUser.loginWithPassword;
     this.requireUser = $meteorUser.requireUser;
--- a/modules/angular-meteor-object.js
+++ b/modules/angular-meteor-object.js
@@ -14,6 +14,11 @@ angularMeteorObject.factory('AngularMeteorObject', ['$q', '$meteorSubscribe', fu
     return this;
   };
 
+  AngularMeteorObject.$meteorSubscribe = function () {
+    this.subscribe.apply(this, arguments);
+    return this;
+  };
+
   AngularMeteorObject.save = function save(docs) {
     var self = this,
       collection = self.$$collection;
@@ -82,7 +87,7 @@ angularMeteorObject.factory('AngularMeteorObject', ['$q', '$meteorSubscribe', fu
 
 // A list of internals properties to not watch for, nor pass to the Document on update and etc.
   AngularMeteorObject.$$internalProps = [
-    'save', 'reset', '$$collection', '$$options', '$$id', '$$hashkey', '$$internalProps', 'subscribe', 'stop', 'autorunComputation', 'unregisterAutoBind', 'unregisterAutoDestroy', 'getRawObject'
+    'save', 'reset', '$$collection', '$$options', '$$id', '$$hashkey', '$$internalProps', 'subscribe', '$meteorSubscribe', 'stop', 'autorunComputation', 'unregisterAutoBind', 'unregisterAutoDestroy', 'getRawObject'
   ];
 
   var createAngularMeteorObject = function(collection, id, options){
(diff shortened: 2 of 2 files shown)


Urigo:
Why is that different from subcribe?

alansikora:
It's not, that just because there was a change from $scope.subscribe to $scope.$meteorSubscribe, and I do have some projects that use a scoped subscribe but falls back to global subscribe if no scope is passed.

As $meteor just have the "subscribe" method, that would need an extra "if" to determine which method is needed to be used to do the same thing.

Doesn't it makes sense to keep methods that do the same thing with the same name?

Urigo:
can you please share your fallback code?
is it something worth adding here as well?
If it's good for object, it can be good for collection as well no?

I'm not sure yet, because the reason we added $meteorSubscribe is because it's on $scope, calling $meteor.$meteorSubscribe is a bit too much no?

alansikora:
Let me show you a simple example:

```javascript
function myCustomSubscribe( $scope, name ) {
  var subscribeArgs = Array.prototype.slice.call( arguments );
  subscribeArgs.splice( 0, 2 ); // remove $scope and name

  var $subscriber = $scope || $meteor;

  $scope.$meteorSubscribe.apply( $scope, subscribeArgs );
}
```

And, without this, I would need to do this:

```javascript
function myCustomSubscribe( $scope, name ) {
  var subscribeArgs = Array.prototype.slice.call( arguments );
  subscribeArgs.splice( 0, 2 ); // remove $scope and name

  if ( $scope ) {
    $scope.$meteorSubscribe.apply( $scope, subscribeArgs );
  } else {
    $meteor.subscribe.apply( $meteor, subscribeArgs );
  }
}
```

It's really not big deal, but as we are dealing with the very same functionality, angular-scoped or not, I think that it would make sense to keep the function names equal.

And yes, I think $meteor.$meteorSubscribe is too much, that's why I think we should keep the $meteor.subscribe method, and use $meteorSubscribe just as an alias, to keep things sane when you need to.

By the way, thank you for your attention, this project is awesome and we are using in real-world projects at my company, expect some more PR's along the way ;)

Urigo:
thanks, that's an interesting scenario.
Just to understand a bit more, why do you need to create a custom subscribe?
I mean, I was thinking that right now, after we have all the nice scope handlers, why would you want to use a different one?

alansikora:
We have a seed project with commonly needed structures, like factories/services, extended from a parent that has methods like subscribe/save/remove, overwritable.

Given that scenario, the subscribe method is normally scoped to the current controller, but sometimes we need that to be global, for any reason.

Urigo:
I'm sorry but it looks like a very specific scenario which also has a solution.
I understand your need and it's a cool thing, but I don't want to bloat the code with another function.

I do look forward for more pull requests, I hope you understand my thinking on this

alansikora:
Of course Uri, I understand your thinking, but as seen in some known libraries like lodash and underscore, aliases are used quite often to keep things sane and programmer friendly. Just my 2c!

Keep up the good work, and if you ever plan to coming to Brazil, I might find a place around ;)

auterium:
I agree with @Urigo, adding more functions even aliases would make it unnecessarely large. If this is a common scenario in your projects, why don't you create it as an angular service part of your seed project?

alansikora:
Well, that's fine folks :)

For me, it makes no sense to have two different methods that do the same thing, at least there should be a way to make it equal. We spent like 30 minutes to find why subscribe was not working on "$scope", and we are used to the project, we had to crawl into the code.

I disagree, honestly, but who am I?

Thank you for your attention!

auterium:
@alansikora I agree that 2 methods that do the same should not have different names. Even further, there should not be 2 methods that do the same thing.

This case is different: $scope.$meteorSubscribe, $scope.$meteorCollection, and $scope.$meteorObject are wrappers that add extra functionality that depends on $scope (they add a listener to $scope.$on('$destroy') to stop the subscription/collection/object), while $meteor.subscribe, $meteor.collection, and $meteor.object are unaware of $scope.

I'm sorry that you had to look for and change all the $scope.subscription usages, and I appologize for that, as it was a method that only existed for 1 minor version, but that's the risk of using cutting edge, still in progress, technologies.

I invite you to keep an eye on issues & PRs in this project to have a deeper understanding on the direction of the project and be part of it's contributors with code and/or arguments (either are always a good help to decide the direction of any project.

Regards,
Marc

alansikora:
@auterium don't think I'll be away for too long :D

I love the project, we are using for a couple of months so far, and you can be sure I'll be in touch.

Urigo:
@alansikora thank you so much for this.
Keep bringing up interesting use cases

PR #344, opened 2015-05-18
Collection rendering speed improvement

barbatus:
Collection rendering speed improvement based on an async $apply to a bulk of changes made to the collection at once.


barbatus:
ng-meteor.meteor.com  was used for manual testing.

PR #352, opened 2015-05-21
mongo documents with ObjectId

Tallyb:
Explanation about retrieving documents that are stored with object Ids


Urigo:
Cool, can you add a code example?

Urigo:
Thanks @Tallyb !

PR #356, opened 2015-05-25
fix code snippet that is not displayed as code

oshai:
I am not sure if this fix is correct - didn't know how to preview it. anyway in the original doc this part is not displayed as code, and something bad with the formatting. I tried to fix indentation.


Urigo:
Thanks!

PR #360, opened 2015-05-26
Update tutorial.step_09.tpl

blevinscm:
Added instruction on line 173 to tell the student to explicitly put the new $scope inside the controller.  Corrected type of Straight and not sure about this since I am a novice but think line 183 should say that we are subscribing to party incase we went straight to the details page so didnt resolve in the router.


Urigo:
Thanks @blevinscm

PR #361, opened 2015-05-26
Update tutorial.step_09.tpl

blevinscm:
I tried to figure out the code to show users that logged in with a service, but I couldn't so until that code can be put in I recommend the above disclaimer.


Urigo:
@blevinscm Good idea.
Can you add the json structure of each type of login?
can help even more

blevinscm:
Added.  I am fairly new to GitHub so not sure if I did correctly. 
Thanks.

Urigo:
Thanks!

PR #371, opened 2015-05-27
Adjusting logic to return Date objects

cantoine:
I made an adjustment to the stripDollarPrefixedKeys so that Date objects are returned instead of being destroyed during parsing. I ran into this problem while working with Simple Schema and a standard Date field. I don't believe this simple change will have any negative effects.

If I've missed any protocols or anything for submitting this change let me know.

Thanks!
Chris

### DIFF
--- a/modules/angular-meteor-utils.js
+++ b/modules/angular-meteor-utils.js
@@ -25,7 +25,7 @@ angularMeteorUtils.service('$meteorUtils', [ '$timeout',
     };
     // Borrowed from angularFire - https://github.com/firebase/angularfire/blob/master/src/utils.js#L445-L454
     this.stripDollarPrefixedKeys = function (data) {
-      if( !angular.isObject(data) || data instanceof File) { return data; }
+      if( !angular.isObject(data) || data instanceof File || data instanceof Date) { return data; }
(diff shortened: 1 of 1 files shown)


Urigo:
thanks @cantoine 
can you please add a test for this scenario?

Urigo:
also another question - if we would change this function into angular.copy, will it work?

cantoine:
I did some quick testing and it appears that angular.copy would work as well, if I'm understanding correctly. Are you thinking of removing this function entirely and just using angular.copy in its place or something else?

I would be happy to make changes and tests, just let me know.

Thanks!
Chris

Urigo:
Yes I'm thinking about removing it...
I've asked someone from angularFire why did they use this instead, waiting for an answer soon, sorry for the delay
Will answer soon

cantoine:
Great, if I can help in some capacity just let me know and thanks for the super quick replies.

Urigo:
@cantoine I think adding tests will be great anyway!

Urigo:
@cantoine I'm closing this but hope to get a test for Date PR

cantoine:
@Urigo sounds great, I'll get a test written and get another PR created.

Thanks!

PR #384, opened 2015-05-29
session-bind

Tallyb:
fix to session.bind api reference to remove template handler and make it pure angular


Tallyb:
wrong files added.

PR #385, opened 2015-05-29
Session bind

Tallyb:
Fix to session.bind api reference to remove template handler and make it pure angular. 

(Readme was not changed - fail attempt to add the definetly typed badge. looks ill).


Urigo:
@Tallyb thank you so much but the whole point here is to show how Meteor's session and session way work directly with angular even if it's not angular updating the variable

Tallyb:
Ah! Well, I could not guess this hidden intention, but maybe it is just me... In case not, I will look into augmenting it with code comments to explain this. The MainCtrl / detailCtrl also has some meaning, or just a typo ?

Urigo:
I think there is no meaning to the detailCtrl

PR #388, opened 2015-05-29
Escape {{ }} in camera example

Brasilikum:
I am not sure how to actually escape `{{party.imageData}}`. It is in the source code, but as the code is being processed, `{{party.imageData}}` does not appear in the [camera API docs](http://angular-meteor.com/api/camera). Please propose fix.
I noticed it because the example did not work for me. Then I looked it up in the video of Urigo and saw the curly braces.


Urigo:
@Brasilikum thanks!  I've fixed it with the help of Blaze helpers.
I will change the rest of the tutorial code to do the same sometime next week

Issue #383, opened 2015-05-29
Inserting a new document


MurWade:
Inside my controller i have 
`$scope.posts  = $meteor.collection(Posts)`

I can insert a document doing it this way

`$scope.posts.pust(newPost)`

or

`Posts.insert(newPost); // the meteor way `.

They both do the same time and binding works and all . doing it the meteor way also returns the id which is very very useful. So i was wondering what is the difference or which is the proper way to insert new document. Thank you.

vacarsu:
Well pushing a new document to a scope variable assumes you have already queried a list of documents. Posts.insert would be useful in cases when you haven't, or don't need to grab a list before inserting a new document. I haven't even gotten this far as I get an error. This is close to the question I asked so I'll reference. #377

vacarsu:
It seems you have answered my question! I didn't realize it unnecessary to return collection.insert from a $meteor.collection callback.

Urigo:
@MurWade 

    $scope.posts.push(newPost)

is for easiness when you use autobind and want to use the regular JS Array API.
If you don't want that, a better practice would be to remove the autobind and use

    $scope.posts.save(newPost)

Which returns a promise on success with the id and the action happened ("inserted" or "updated"):
https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-meteorCollection.js#L46

Do you think we should document this somewhere?
if so, where and maybe you would like to pull request the docs for that?

PR #397, opened 2015-06-02
Angular Meteor Collection API doc

Tallyb:
Added the return value for the save as discussed in https://github.com/Urigo/angular-meteor/issues/383


Urigo:
Thanks @Tallyb !

PR #404, opened 2015-06-04
Return params for remove

Tallyb:


Urigo:
Thanks @Tallyb , also nice issue number...

Tallyb:
probably because the previous one failed on authorization problem ;)

PR #419, opened 2015-06-08
new logo

Tallyb:
new logo to angular meteor website


Urigo:
@Tallyb can you increase the side a bit and also make it look unpixelated?
thanks!

Tallyb:
@Urigo it probably looks different on every screen. I am also not sure what side (can be also done with padding?). I have created another PR to add the file to the logos file, so you can take it further as it is seen on your computer. It is a simple SVG (no fancy ai, psd...)

Issue #418, opened 2015-06-08
Date is interpreted as an object in diff-array.js


bjarketrux:
I have a problem of a date not being updated on the database when a client changes the value (value is updated on client).

The code in `diff-array.js` looks like it interprets dates as an objects, and tries to compare all properties on it, but there isn't any.

```
// Can tell whether a value is an object and not an array
var isActualObject = function (value) {
  return angular.isObject(value) && !angular.isArray(value);
};
```

If you call the above code with a `new Date()` as the value it will return true.
Is this a bug or intended behavior?

Thanks

Urigo:
@bjarketrux do you think you can add a failed test that catches this scenario?

bjarketrux:
First thing tomorrow!

bjarketrux:
Fixed in pull request #422

PR #422, opened 2015-06-09
Pull request for issue #418 - Date is interpreted as an object in diff-array.js

bjarketrux:
Fixed a bug where different dates did not trigger an "is changed" updated.
Issue #418


Urigo:
Great @bjarketrux !
Can you confirm that the test failed before the fix?

bjarketrux:
Yes I did it the school-boy way ;)
I'll close the bug.

PR #431, opened 2015-06-12
Small fix on error log message

ccortezia:
Was referencing $meteorCollection. 
Probably a copy+paste typo.


Urigo:
thanks @ccortezia !

PR #438, opened 2015-06-16
Cfs file insert fix

mxab:
A common use case is that you insert FS.File instances in a cfs collection, the strip hash key function ignored this case and returns a object. This results in an error if you do something like
```
var fsFile = new FS.File(file)
fsFile.ownerId = Meteor.userId();
MyCollectionFS.insert(fsFile);
```

I added the handling of FS.File to the strip function


Urigo:
Very nice!
can you make the if statement a bit nicer, something like

    if (something1 ||
        something2 ||
        (something3 && something4))

?

Urigo:
I'm merging, but can you add another pull request after?
thanks

PR #443, opened 2015-06-17
unify allow code

dotob:


Urigo:
thanks @dotob but I wanted to leave it like that so it will be more obvious for people what's actually happening....
what do you think?

dotob:
i just thought that the check for update and remove is the same as for insert, but they had different code. thats what i wanted to unify. or did i misunderstood the insert vs update/remove?

Urigo:
yes I think you are right. also @netanelgilad convinced me :)   thanks!!

PR #458, opened 2015-06-22
Trivial grammar fix step 9

ppotoplyak:
service that also ``return`` -> service that also ``returns``


Urigo:
Thanks @ppotoplyak !

PR #466, opened 2015-06-24
added getCollectionReactively to scope

mxab:
Hi,
I added a method to scope that to support angular $watchCollection as it should provide a more performant way if you want to react on array changes in terms of pushes etc than using ```$scope.getReactively("myValues", true)``` way


Urigo:
Great!
can you add some tests?

mxab:
ok, have a look

Urigo:
Great stuff @mxab !

Urigo:
@mxab can you add it also to the docs please?
thanks

mxab:
you are so demanding ;) can you check if this is ok?

Urigo:
I am am I ? ;)
thank you so much!

PR #467, opened 2015-06-24
Add $meteor.collectionFS for handeling these collections

dotansimha:
- Added ability to override the "diffArray" of the $meteorCollection base object in case of collection with special objects (like CollectionFS).
- This PR also the solution for "Max Call Stack" errors when trying to work with CollectionFS.


modules/angular-meteor-meteorCollection.js
```diff
@@ -231,10 +231,10 @@ angularMeteorCollections.factory('AngularMeteorCollection', ['$q', '$meteorSubsc
       }
     };
 
-    var createAngularMeteorCollection = function (cursor, collection) {
+    var createAngularMeteorCollection = function (cursor, collection, diffArrayFunc) {
       var data = [];
       data._serverBackup = [];
-
+      data.diffArrayFunc = diffArrayFunc;
```
Urigo: @dotansimha why?

Urigo:
Thanks @dotansimha

PR #473, opened 2015-06-26
Update tutorial.step_02.html

sorahavok:
Not sure what the original intent of this part was but there was some data missing and it was confusing.


Urigo:
Thanks for the catch @sorahavok , I've made a mistake a few commits ago about that.
I've fixed that a bit differently to return to the original content.

PR #475, opened 2015-06-26
Fix bad behavior of diffArray on the presence of nulled and empty objects 

### LINKED ISSUE #469: Problems setting nested object using AngularMeteorCollection.save
Suppose I have an object as simple as ```{brand: 1, value: {sub: 2}}```

After setting the field "value" to null through AngularMeteorCollection.save(data), the local collection no longer updates itself to any posterior non-null values on the "value" field (altough the subscription is properly receiving the changes). The field "brand" is normally update though.

PS: I detected this behavior when preparing a sandbox to debug another related problem. In the other scenario, I had a stack trace accusing TypeError inside the function ```diffObjectChanges``` on the [line](https://github.com/Urigo/angular-meteor/blob/master/lib/diff-array.js#L149) ```var diff = diffObjectChanges(oldItem[key], value);``` telling ```Exception in queued task: TypeError: Cannot read property 'blah' of null```. Unfortunately I couldn't reproduce this stacktrace on the sandbox, but the steps I used inside my bigger project are just the same.

The sandbox repo is linked [here](https://github.com/ccortezia/angular-meteor-sandbox/tree/bug-nested-diff)

ccortezia:
fixes #469

### DIFF
--- a/lib/diff-array.js
+++ b/lib/diff-array.js
@@ -116,7 +116,7 @@ var flattenObject = function (object, parentKey) {
   var flattened = {};
 
   angular.forEach(object, function (value, key) {
-    if (isActualObject(value)) {
+    if (isActualObject(value) && !_.isEmpty(value)) {
       angular.extend(flattened, flattenObject(value, key));
     } else {
       var dotNotedKey = (parentKey) ? parentKey + "." + key : key;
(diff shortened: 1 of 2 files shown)


Urigo:
@ccortezia this is awesome change!!

PR #478, opened 2015-06-30
Added disclamer to authentication

westlakem:
Added a disclaimer that states authentication on the client (via ui-router) can be circumvented by someone with malicious intent.


Urigo:
@westlakem good point.

Can you just add right after that this is the reason we also made restrictions on the server (just so users will understand that they've done what they need and that the step in complete and secure)

westlakem:
@Urigo It's done.. Travis CI is still building, but the extra piece is in there.

PR #498, opened 2015-07-05
Fix cancel button syntax problem.

oneonestar:
Buttons should not contain any interactive descendants.
"<button><a href="/parties">Cancel</a></button>" is invalid in HTML5.
This doesn't work in Firefox 38.

Reference: https://stackoverflow.com/questions/2906582/how-to-create-an-html-button-that-acts-like-a-link

### DIFF
--- a/.docs/angular-meteor/client/views/steps/tutorial.step_06.html
+++ b/.docs/angular-meteor/client/views/steps/tutorial.step_06.html
@@ -82,7 +82,7 @@ __party-details.ng.html__
 
     <button ng-click="save()">Save</button>
     <button ng-click="reset()">Reset form</button>
-    <button><a href="/parties">Cancel</a></button>
+    <button onclick="window.location.href='/parties'">Cancel</button>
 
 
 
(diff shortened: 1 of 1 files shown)


Urigo:
Thanks @oneonestar 
so maybe it's better to change it to:

    <button ng-click="window.location.href='/parties'">Cancel</button>

?

oneonestar:
Inside ng-click is an AngularJS expression.
Angular expressions do not have access to global variables like window, document or location. [1](https://docs.angularjs.org/guide/expression)

The standard way for AngularJS to navigate page is to write a function and deal with the $location stuff. I think using native javascript here is alright, keeping the tutorial workable and concise.

Tallyb:
ng-click="$state.go ('parties')"

oneonestar:
$state.go() is used in a controller and ui-sref is used in HTML. [1](http://stackoverflow.com/questions/24526801/difference-between-ui-sref-and-state-go-in-angularjs-ui-router)
Now changed to:
```    
    <button ui-sref="parties">Cancel</button>
```
Seems working working pretty fine in Firefox and Chrome.

Urigo:
Good stuff!

PR #501, opened 2015-07-06
Fixed deeplinks in the docs that weren't working

markleusink:
using angular-meteor.com instead of angularjs.meteor.com)


Urigo:
Too many pull requests :)
now it's outdated again :)
but this time it will probably be easier to update.
Thanks and sorry it was so hard to contribute

Urigo:
Thanks!

Issue #503, opened 2015-07-06
Problem when trying to $set object to nulled field


ccortezia:
Suppose I have a record ```{nested: null}```. If I try to update it to ```{nested: {sub: 1}}```, I get an exception. The record is properly updated though.

The execution of this [deepCopyObjectChanges](https://github.com/Urigo/angular-meteor/blob/master/lib/diff-array.js#L246) raises the following exception: ```Exception in queued task: TypeError: Cannot set property 'sub' of null```

Isolated repo [here](https://github.com/ccortezia/angular-meteor-sandbox/tree/bug-nested-diff-2). 
Click on clear,save: bug appears. 
Click on drop,save: bug won't appear.

Urigo:
@ccortezia it is closed by your pull request?

ccortezia:
Yes, we can close it now.

PR #506, opened 2015-07-07
Fix meteorCollection bad behavior with deep-objects update and null fields

ccortezia:
Several different combinations of "new-field-type" and "previous-field-type" led to broken behavior on meteorCollection. Having null as a previous value on a field, for instance, could break if the field was both at a shallow or a deep position in the object. Big problem is the real collection data is updated and persisted, but the angular collection sync code would raise a TypeError some time after that, and potentially affect angular execution. 

This PR assures good behavior on the scenarios described by the added tests, but the sad reality is that dealing with object-diff and read-write of deep fields is challenging. I think we should consider bringing pieces of more stable code into angular-meteor in order to improve diff-array.js. 

For example, lodash ```_.get```, ```_.set``` and ```_.has``` may help us a great deal. A stripped version of the library with only these three functions will add something around 3KB to the minified footprint.

### DIFF
--- a/lib/diff-array.js
+++ b/lib/diff-array.js
@@ -225,7 +225,8 @@ var handleDeepProperties = function(object, paths, cb) {
     var currentObject = object;
 
     for (var i = 0; i < props.length - 1; i++) {
-      if (!_.isObject(currentObject[props[i]]))
+      var tmp = currentObject[props[i]];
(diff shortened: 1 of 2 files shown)


ccortezia:
Relates to #503

Urigo:
@ccortezia great points and great pull request.
I would love to hear more about changing our own diff into an existing one

PR #508, opened 2015-07-09
Fix ResetCtrl function name, some typos

tschaei:
In ResetCtrl the function was mistakenly named register instead of reset. This does'n match the following template code which expects the name to be reset


Urigo:
Thanks @tschaei !

PR #515, opened 2015-07-14
Might be better ?

ivanproskuryakov:


Urigo:
@ivanproskuryakov thanks for the suggestion but with your change, it will remove the parties on every load....
We want to do that only in case it's the first time you load the app and there are not existing parties..

PR #529, opened 2015-07-21
Added a section to the docs for the ng.js and ng.html file extensions

markleusink:
That info needs to be in the docs too - not only hidden away in the tutorial ;)


Urigo:
that's awesome!

PR #530, opened 2015-07-22
Update mongo-collection-instances

rclai:
This update fixes a couple issues regarding `Mongo.Collection` wrapping.


Urigo:
thanks @rclai !

PR #543, opened 2015-07-27
Add bundle-minified script and update bower dependencies

idanwe:
* Create bundle and minified file with the meteor dependencies
 * Add angular as bower dependency
 * Add meteor-client-side as bower dependency

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "angular-meteor",
-  "main": "angular-meteor.js",
+  "main": ".dist/angular-meteor.bundle.min.js",
   "version": "0.9.3",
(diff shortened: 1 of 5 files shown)


bower.json
```diff
@@ -34,5 +34,9 @@
     "package.js",
     "package.json",
     "versions.json"
-  ]
+  ],
```
Urigo: Make sure you have exactly the files you need

dist/bundler.sh
```diff
@@ -0,0 +1,41 @@
+cd ../..
```
Urigo: Move all bower folders to .folder

PR #549, opened 2015-07-28
angular 2 tutorial. v1

ShMcK:


ShMcK:
Tutorial updated. Important fixes.

PR #556, opened 2015-07-31
Use angular $parse to handle nested mode (a.b.value)

topikachu:
MeteorSession doesn't work with nested mode something link 'a.b.c'
Fix it by using angular $parse


Urigo:
@topikachu thanks!
can you please add some unit tests for those scenarios?

topikachu:
@Urigo  done! Please check the new change.

Urigo:
thank you so much!

PR #568, opened 2015-08-05
Corrects prev changes that didn't get into the code base.

barbatus:
Removes async tests since they are not needed.
Adds one test to test removing all docs.
Deployed along with ToDo app here http://ng-to-do-test.meteor.com

### DIFF
--- a/modules/angular-meteor-collection.js
+++ b/modules/angular-meteor-collection.js
@@ -6,7 +6,8 @@ var angularMeteorCollection = angular.module('angular-meteor.collection',
   ['angular-meteor.stopper', 'angular-meteor.subscribe', 'angular-meteor.utils', 'diffArray']);
 
 // The reason angular meteor collection is a factory function and not something
-// that inherit from array comes from here: http://perfectionkills.com/how-ecmascript-5-still-does-not-allow-to-subclass-an-array/
+// that inherit from array comes from here:
+// http://perfectionkills.com/how-ecmascript-5-still-does-not-allow-to-subclass-an-array/
 // We went with the direct extensions approach
 angularMeteorCollection.factory('AngularMeteorCollection', [
   '$q', '$meteorSubscribe', '$meteorUtils', '$rootScope', '$timeout', 'diffArray',
@@ -22,7 +23,7 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
       if (angular.isDefined(collection))
         data.$$collection = collection;
       else
-        data.$$collection =  $meteorUtils.getCollectionByName(cursor.collection.name);
+        data.$$collection = $meteorUtils.getCollectionByName(cursor.collection.name);
 
       angular.extend(data, AngularMeteorCollection);
       return data;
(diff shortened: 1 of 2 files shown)


Urigo:
@DAB0mB

modules/angular-meteor-collection.js
```diff
@@ -53,14 +54,61 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
       return allPromise;
     };
 
-    AngularMeteorCollection.remove = function (keys) {
-      // remove collection
-      if (!keys) _.pluck(this, '_id');
-      // remove single doc
-      keys = [].concat(keys);
+    AngularMeteorCollection._upsertDoc = function(doc, useUnsetModifier) {
```
DAB0mB: https://github.com/barbatus/angular-meteor/blob/master/modules/angular-meteor-object.js#L40
Change function name as well

barbatus: @DAB0mB Hey, what kind of collection is it supposed to be here 
https://github.com/barbatus/angular-meteor/blob/master/modules/angular-meteor-object.js#L21?
Mongo.Collection doesn't have upsertObejct method at all.

DAB0mB: You're right, it should be `AngularMeteorCollection._upsertDoc.call(this, updates)`. I'm currently checking $meteorObject with a small sample app I built and I actually had some problems with that

barbatus: @DAB0mB But why _upsertDoc?  AngularMeteorCollection's ````save```` does same thing.
I think here https://github.com/barbatus/angular-meteor/blob/master/modules/angular-meteor-object.js#L21 should be ````$meteor.collection(collection)```` and then at the object's save method correct to ````$$collection.save(updates)````

DAB0mB: I agree, it does look more "clean". And also I found this https://github.com/barbatus/angular-meteor/blob/master/modules/angular-meteor-object.js#L139. `collection` is not defined, it would be nice if you could fix this in your pull request. As I said I'm currently checking my changes, but I also think it would be a good idea to add tests to `$meteorObject`.

Urigo:
@barbatus @DAB0mB what is the status on your end?
Is it ready for merge?

barbatus:
@Urigo While I am ok to fix this,  @DAB0mB is welcome to write multiple tests since those bugs in the  $meteorObject and not only like this one https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-collection.js#L58 were induced by his refactoring.

DAB0mB:
@Urigo @barbatus I'm currently writing some tests to `$meteorObject` and fixing some bugs. Apparently the tests are not covering every scenario and I was relying on them too much without manually checking my changes.

DAB0mB:
@barbatus I made a pull request that fixes all the problems (atleast what I noticed). See pull request #571. I suggest to first merge your pull request, this time it will be alot easier to deal with conflicts. If youv'e found any other bugs, do tell me so I can fix them in my pull request, and also your'e more than welcome to comment on my changes :-)

PR #569, opened 2015-08-05
Update tutorialAngular2.step_04.html

freezby:
1) Missing client folder in template url (does not break the app but may be more consistent with previous steps ?)
2) Wrong name for the form controller (partyForm instead of partiesForm)


Urigo:
@ShMcK

ShMcK:
Thanks! Looks good.

PR #574, opened 2015-08-08
Update tutorialAngular2.step_05.html

PierrePIRONIN:
Method "remove(party) {...}" was not moved from app.ts to parties-list.ts so delete party button is broken


Urigo:
Thanks

PR #575, opened 2015-08-08
Minor fixes in tutorialAngular2.step_06.html

PierrePIRONIN:
- Remove useless Meteor.isClient test because we are in client directory and remove too many closing braces at the end
- Make router-link '/parties-list' homogeneous with the beginning of document


Urigo:
great!

PR #581, opened 2015-08-12
fix #528

### LINKED ISSUE #528: getReactively returns no value in initial run
Hi,
if I init a variable with initial default value and then use getReactively in a reactive function the value is not returned

for example
```
vm.lang = 'en';
vm.messages = $scope.$meteorCollection(function () {
        var lang = $scope.getReactively('vm.lang');
        if(!lang){
            console.warn('lang not set', lang);
            lang = 'en';
        }
        return Messages.find({
            lang : lang
        });
    }, false);
```

maybe the reactive function should run after the first digest I guess

dszczyt:


Urigo:
@dszczyt very nice.
But how would you prevent the original reason of the check to happen?
also, can you add tests that cover issue 528?
thanks!

dszczyt:
Don't merge... It breaks other things, and I don't know why. My original issue was : in a controller, when we run "subscribe" with "scope.getReactively" parameters initialized juste before, subscription doesn't "see" the values in the scope. Wrapped it with a $timeout and it works. Ugly but efficient. I will try to write a test case for this issue. Thanks.

Urigo:
@dszczyt ok, waiting for your test so we can have a closer look.
Thanks

Issue #483, opened 2015-07-02
Collection2 errors will result in "$apply already in progress" error


shadysayed:
In a project that uses collection2 for schema validation. saving an invalid document using

    $scope.collection.save(newDoument)

will result in "$apply already in progress" the reason seems to be that you call $rootScope.$apply without first checking $rootScope.$$phase inside AngularMeteorCollection.save function. And for some reason the callback for inserting an item in a collection is called inside a digest phase.

to avoid this I usually create a helper that goes like this

    $rootScope.applyIfNeeded = function(callback) {
      if (this.$$phase) {
        if (typeof callback === "function") {
          callback();
        }
      } else {
        this.$apply(callback);
      }
    };

Then whenever I need to call $apply I use

    $rootScope.applyIfNeeded(<callback if any>)

shadysayed:
I've created a repo to reproduce this issue. Check out this commit shadysayed/meteor-issues-reproduction@1566e2e91ffbe17c3e13fda5925445978e2b7f74

Urigo:
@shadysayed thanks for noticing!
do you care of creating a pull request with a fix and a failing (before) and success (after) tests?

Urigo:
@shadysayed can you please test this issue against the latest master?

cranesandcaff:
I'm on `0.9.3`, it's still throwing on incorrect docs.

DAB0mB:
As for version `1.0.0` there isn't any digestion problem. Obviously the promise won't be resolved since it's being rejected by the schema, but the funny thing is that the promise is rejected with the bound result instead of the received error. I [fixed](https://github.com/DAB0mB/angular-meteor/commit/f0e6a71f5625f916b98bf92870a59bab48c0edf5) it, hopefully enough it will be merged soon :-D

Urigo:
@shadysayed looks like it's fixed in `angular@1.0.0-rc.5`.
Please test and let me know if it works for you.

markleusink:
Latest version fixed it for me!

(after updating the package name from urigo:angular to angular ;-)

PR #582, opened 2015-08-13
One more pass of meteorCollection re-factoring.

barbatus:
Main purpose is to clearly separate server update logic from the client one.
So that we don’t have any special variable that sets mode globally (like "UPDATING_FROM_SERVER")
For each update from the server or any internal changes to the collection, we perform
_unsetAutoClientSave
… here changes go
_setAutoClientSave

Also ````collectionUtils```` has been changed to provide collection with the changes only,
saving of the changes collection take care of by itself from now own, which simplified code flow.

All Tracker.Autotuns now are stopped on the collection’s stop method as well.

Also, I seem to have understood the reason of https://github.com/meteor/simple-todos-angular/issues/2.
It was very likely due to uncontrollable ````setAutoBind````, which may have run
multiple time here https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-collection.js#L284, thus, ended up having multiple ````$watches```` at the same time.

#483 issue looks like fixed already.

ToDo app with the current commit deployed here http://ng-to-do-test.meteor.com.

angular:angular-mocks rolled back to the previous version; new version seems to require some new dependancies.

We also certainly need to devote some time to go through all the current bugs (particularly for meteorCollection) and fix them before the next version. @Urigo what do you think?
Also, how about to set up some code style guide to follow? Code files look sometimes too different to my mind.

### DIFF
--- a/modules/angular-meteor-collection.js
+++ b/modules/angular-meteor-collection.js
@@ -275,29 +275,43 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
 
     AngularMeteorCollection._saveChanges = function(changes) {
       // First applies changes to the collection itself.
+      var newDocs = [];
       for (var itemInd = changes.added.length - 1; itemInd >= 0; itemInd--) {
         this.splice(changes.added[itemInd].index, 1);
+        newDocs.push(changes.added[itemInd].item);
+      }
+      // Then saves all new docs in bulk.
+      if (newDocs.length) {
+        this.save(newDocs);
       }
 
-      // Then saves additions.
-      for (var itemInd = 0; itemInd < changes.added.length; itemInd++) {
-        this.save(changes.added[itemInd].item);
-      }
-
-      // Then saves removes.
+      // Collects docs to remove.
+      var removeDocs = [];
       for (var itemInd = 0; itemInd < changes.removed.length; itemInd++) {
-        this.remove(changes.removed[itemInd].item);
+        removeDocs.push(changes.removed[itemInd].item);
+      }
+      // Removes docs in bulk.
+      if (removeDocs.length) {
+        this.remove(removeDocs);
       }
 
-      // Then saves changes.
+      // Collects set and unset changes.
+      var setDocs = [], unsetDocs = [];
       for (var itemInd = 0; itemInd < changes.changed.length; itemInd++) {
         var change = changes.changed[itemInd];
         if (change.setDiff) {
-          this.save(change.setDiff);
+          setDocs.push(change.setDiff);
         }
         if (change.unsetDiff) {
-          this.save(change.unsetDiff, true);
+          unsetDocs.push(change.unsetDiff);
         }
+      }
+      // Then saves all changes in bulk.
+      if (setDocs.length) {
+        this.save(setDocs);
+      }
+      if (unsetDocs.length) {
+        this.save(unsetDocs, true);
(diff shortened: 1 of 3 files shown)


modules/angular-meteor-collection.js
```diff
@@ -184,87 +231,107 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
...
+        self._setAutoClientSave();
+      }, true);
     };
 
-    AngularMeteorCollection.setAutoBind = function() {
-      this._isAutoBind = true;
+    AngularMeteorCollection._saveChanges = function(changes) {
+      // First applies changes to the collection itself.
+      for (var itemInd = changes.added.length - 1; itemInd >= 0; itemInd--) {
+        this.splice(changes.added[itemInd].index, 1);
```
Urigo: When applying added items, you have to also remove them from the local collection, in case they are not in the mongo filter

modules/angular-meteor-collection.js
```diff
@@ -184,87 +231,107 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
...
+    };
 
-      if (this.observeHandle)
-        this.observeHandle.stop();
+    AngularMeteorCollection._setAutoClientSave = function() {
+      var self = this;
 
-      this.splice(0);
-      this._serverBackup.splice(0);
+      // Always unsets auto save to keep only one $watch handler.
```
Urigo: Have you tested against the scenario of which I've added the _isAutobind flag?  this is a complex scenario

modules/angular-meteor-collection.js
```diff
@@ -11,36 +9,58 @@ var angularMeteorCollection = angular.module('angular-meteor.collection',
...
     var deepCopyChanges = diffArray.deepCopyChanges;
     var deepCopyRemovals = diffArray.deepCopyRemovals;
 
-    function AngularMeteorCollection (cursor, collection, diffArrayFunc) {
+    function AngularMeteorCollection(curDefFunc, collection, diffArrayFunc, autoClientSave) {
       var data = [];
       data._serverBackup = [];
-      data.diffArrayFunc = diffArrayFunc;
+      data._diffArrayFunc = diffArrayFunc;
+      data._hObserve = null;
```
Urigo: Can you add comments about each of those variables and what their jobs are?

Urigo:
@barbatus can you rebase your changes please?

modules/angular-meteor-collection.js
Urigo: ohh sorry, it looks like you do that, so in that case, a better comment will be good

modules/angular-meteor-collection.js
barbatus: Yes, I've tested it against scenario defined in the bug https://github.com/meteor/simple-todos-angular/issues/2.
As mentioned, I came across similar situation when doing that PR: as I could see it's likely related to this line https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-collection.js#L284 which could be executed twice in a row therefor leaving one $watch handler uncontrollable.

modules/angular-meteor-collection.js
barbatus: done

Urigo:
@barbatus thanks.
I think we should add a test for pushing docs that are not in the mongo filter

PR #588, opened 2015-08-16
docs(tutorialAngular2.step_05.html): fix indentation

IgorMinar:


Urigo:
Thanks!
Fixed in #574

PR #591, opened 2015-08-16
Part not very clear

freezby:
The modified part is not very clear. I propose to put the complete code to avoid any mistake.


Urigo:
thanks, good idea

PR #600, opened 2015-08-18
Update tutorial.step_20.html

okland:
Fixed broken saveCroppedImage function.
Added css so image crop is visible,


Urigo:
@okland can you rebase after the new tutorial tool change?

PR #609, opened 2015-08-21
docs(tutorialAngular2.step_05.html): fix typos

IgorMinar:


ShMcK:
@urigo please also accept these clarifications.

PR #637, opened 2015-08-29
Fix idStringify for recent versions of Meteor

leoetlino:
Since commit meteor/meteor@b3096e93, MongoId and the _idStringify
(or _idParse) method is in a separate package.

diffArray currently assumes the _idStringify and _idParse methods to
be in LocalCollection, but this isn't true anymore in recent versions
of Meteor. Therefore, diffArray() throws an error on (for example)
Meteor 1.2.

This commit fixes this issue by using the mongo-id package if we cannot
find LocalCollection._idStringify and LocalCollection._idParse are
undefined.

### DIFF
--- a/lib/diff-array.js
+++ b/lib/diff-array.js
@@ -4,8 +4,8 @@ var module = angular.module('diffArray', ['getUpdates']);
 
 module.factory('diffArray', ['getUpdates',
   function(getUpdates) {
-    var idStringify = LocalCollection._idStringify;
-    var idParse = LocalCollection._idParse;
+    var idStringify = LocalCollection._idStringify || Package['mongo-id'].MongoID.idStringify;
+    var idParse = LocalCollection._idParse || Package['mongo-id'].MongoID.idParse;
 
     // Calculates the differences between `lastSeqArray` and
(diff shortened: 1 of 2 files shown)


Urigo:
Very nice!

leoetlino:
Thanks for merging! :) Is there a way you could merge this and publish a new version on Atmosphere? I actually ran into this bug while trying to update to Meteor 1.2's latest RC and using the new `meteor:angular` package.

(There's also another bug which partly breaks collections, see PR #638.)

leoetlino:
In the meantime, I have added the following to my `libs/init.ng.js` file:

```
if (Package['mongo-id']) {
    Package.minimongo.LocalCollection._idStringify = Package['mongo-id'].MongoID.idStringify;
    Package.minimongo.LocalCollection._idParse = Package['mongo-id'].MongoID.idParse;
}
```

… but I cannot patch the bug in #638 externally, and that prevents me from updating angular-meteor.

Urigo:
@leoetlino just released a new version with your fixes `1.0.0-rc.3`
let me know if it's all good

leoetlino:
Thank you! That one works perfectly fine − no errors at all in the console. :)

PR #638, opened 2015-08-29
Collection: Fix wrong variable in movedTo()

leoetlino:
Fix a wrong variable being used in the movedTo() observe function.
This fixes a regression introduced by a recent refactor (commit 88b50bf).


Urigo:
That's great!  I've seen that and was about to look into it!

Urigo:
@leoetlino can you please add tests to cover that scenario ?

leoetlino:
I'm not actually _that_ familiar with angular-meteor's codebase and tests, so I am unable to do that. Sorry.

Maybe there should be a test to make sure that no error was thrown?

PR #667, opened 2015-09-11
Update tutorial.step_03.md

mweitzel2005:
Shouldn't the app.js references be to the (in the previous step) updated app.ng.js? That would also remove the DI square brackets from the app(.ng).js code samples.

The DiffBox'es would need to be pointing to the correct version though as well.


Urigo:
@mweitzel2005 you are right, but then we should update the whole tutorial for that, including https://github.com/Urigo/meteor-angular-socially itself..

Are you up for it??

dotansimha:
@Urigo @mweitzel2005 
Done! :)

Urigo:
Thanks @dotansimha

mweitzel2005:
Nice one! Thanks @dotansimha.

PR #670, opened 2015-09-11
Update tutorial.step_07.md

stanusk:
Simple cut & paste as suggested in the tutorial without changing the url paths breaks the app (angular loaded repeatedly error). Took me 30 minutes to find the bug.


stanusk:
Ok, stupid me.. Sorry

PR #695, opened 2015-09-20
changed prefix in comment

simonv3:
It should be a reference to the user's meteor developer account name, no?

### DIFF
--- a/plugin/handler.js
+++ b/plugin/handler.js
@@ -14,7 +14,8 @@ Plugin.registerSourceHandler('ng.html', {
 
   // Build the templateCache prefix using the package name
   // In case the template is not from a package but the user's app there will be no prefix - client/views/my-template.ng.html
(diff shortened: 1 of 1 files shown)


Urigo:
@simonv3 I think that depends if it's a local package or a published one but I'm not sure..
Can you make sure and post it here?

simonv3:
@Urigo Hmm, you're right, they are different, but if it's a local package it doesn't need the `my-app` bit at all, then you just do `my-package`. I guess what's different is that the package name of a published package just includes the developer name - ie `simonv3:simon-package` would be the full name of the package, and the `:` would just get replaced with a `_`.

Urigo:
so maybe can you edit it to be a bit more understandable, like saying package name and give an example?

Urigo:
thanks!

Issue #686, opened 2015-09-16
Collection Save no longer returns ID of inserted object


jonmc12:
Problem:
```javascript
$scope.Todos.save(newTodo).then(function(response) {
    // response used to return: [newTodo._id]
    //
    // response now returns: [{
    //     id: undefined,
    //     action: "inserted"
    // }];
}
```

I corrected this locally by updating [angular-meteor-collection.js, line 85](https://github.com/Urigo/angular-meteor/blob/master/modules/angular-meteor-collection.js#L85) from:

```javascript
var createFulfill = _.partial($meteorUtils.fulfill, deferred, null);
```
to:
```javascript
var createFulfill = _.partial($meteorUtils.fulfill, deferred, null, null);
```
I'm not sure if this fix is comprehensive enough for pull request, and probably needs a test added to spec, but has solved the issue for me temporarily.

Urigo:
@jonmc12 would you want to create a pull request with added tests??  
That could be a first step and then I can review it and add notes

josemejiasji:
+1 on this.

When I change the line you suggested, promise returns _id but not the action.

jonmc12:
@Urigo I've submitted a pull request, would welcome any notes

thuanmaster:
+1 @jonmc12  I changed directly in `angular.js` file in .meteor packages and it works as normally.. Many thanks!

Urigo:
Fixed in `1.0.0-rc.10` , thanks to @jonmc12

PR #696, opened 2015-09-20
functionality and spec for passing a function as bound result to mete…

jonmc12:
…orUtils.fulfill

Proposed fix for #686 to return ID of object inserted into a collection.  Retains intention of code to return both ID and action, and puts a basic spec in place for ```$meteorUtils.fulfill```

```javascript
$scope.Todos.save(newTodo).then(function(response) {
    // response now returns: [{
    //     id: "<_id>",
    //     action: "inserted"
    // }];
}
```

### DIFF
--- a/modules/angular-meteor-collection.js
+++ b/modules/angular-meteor-collection.js
@@ -82,8 +82,10 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
     AngularMeteorCollection._upsertDoc = function(doc, useUnsetModifier) {
       var deferred = $q.defer();
       var collection = this.$$collection;
-      var createFulfill = _.partial($meteorUtils.fulfill, deferred, null);
-      var fulfill;
+      var upsertResult = function(action, _id) {
+        return {_id: _id, action: action }
+      }
+      var fulfill, createFulfill;
 
       // delete $$hashkey
       doc = $meteorUtils.stripDollarPrefixedKeys(doc);
@@ -96,12 +98,14 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
         // it can be $set using update.
         delete doc._id;
         var modifier = useUnsetModifier ? {$unset: doc} : {$set: doc};
-        fulfill = createFulfill({_id: docId, action: 'updated'});
+        createFulfill = _.partial(upsertResult, 'updated');
+        fulfill = $meteorUtils.fulfill(deferred, null, createFulfill);
         // NOTE: do not use #upsert() method, since it does not exist in some collections
         collection.update(docId, modifier, fulfill);
       // insert
       } else {
-        fulfill = createFulfill({_id: docId, action: 'inserted'});
+        createFulfill = _.partial(upsertResult, 'inserted');
+        fulfill = $meteorUtils.fulfill(deferred, null, createFulfill);
         collection.insert(doc, fulfill);
       }
 
--- a/modules/angular-meteor-utils.js
+++ b/modules/angular-meteor-utils.js
@@ -45,6 +45,8 @@ angularMeteorUtils.service('$meteorUtils', [
       return function(err, result) {
         if (err)
           deferred.reject(boundError == null ? err : boundError);
+        else if (typeof boundResult == "function")
+          deferred.resolve(boundResult == null ? result : boundResult(result));
         else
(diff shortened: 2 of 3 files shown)


lungkuteo:
My error same it. It don't return _id
My respone
```
[
    {
        _id: undefined
        action: "inserted"
    }
]
```
How to fix it?

jonmc12:
@lungkuteo if you are using jonmc12/angular-meteor it is tested to return the ID.

Urigo:
@barbatus So, my approach is now that we should be as similar to the Meteor API as possible, meaning that is we want to add something, it should probably go into Meteor's API and not just in the Angular package.
In Meteor's `insert` we don't get the action back as we have to know what we are doing, I believe that here when we do `save` we also need to know what action are we doing and not create any logic based on that.
That being said, it works now like that, with the actions and it's just a bug of the id, so I think I will merge that and separately later on will change the API.
Why do you think it was without the actions before?

barbatus:
@Urigo well it was mentioned in the #686 that it had returned ID before. My point was the user always knows what he saves i.e. whether it's new document or old one, so action maybe not needed.

Urigo:
@barbatus yes I agree with you on that, I would just think that we should remove it in a separate pull request

Urigo:
and then maybe even deprecate save in favor of insert, update and such...

Urigo:
and pull request Meteor for a save function

but one step at a time

Urigo:
@jonmc12 can you add also a test for `$meteorCollection` that we get the wanted result after `save`?

Urigo:
@jonmc12 I'm merging but please open another pull request to add a test.
Thanks!
@barbatus I'm with on that, but we will address that after 1.0.0

jonmc12:
@Urigo - glad bugfix worked out, and I'd be happy to add another test.  To be honest, I researched this test on the first pass and was having trouble finding a meaningful convention to test the return value of upsert via Jasmine.  I'll revisit this weekend, if you happen to know of a similar example of such a test, it  would be helpful.

Urigo:
there are a few examples in the tests page

jonmc12:
@Urigo a more specific example would be helpful.  I spent some more time researching, and you are requesting to test the return value of a fulfilled promise, but there is nothing in the code base quite like this as far as I can tell.

I wanted to use an approach like below, but my spy never gets called. 

```javascript
it('should return an _id and action upon insert', function() {
  spyOn($scope.collection, 'save').and.callThrough();
  
  var doc = {_id: 'new-id'},
    successSpy = jasmine.createSpy('success')
  
  $scope.collection.save(doc).then(successSpy);
  
  expect(successSpy).toHaveBeenCalledWith({ _id: "test", action: "inserted"});
  $rootScope.$apply();
});
```

I tried with a call-through on the ```then``` method as well, but the result is the same.

```javascript
it('should return an _id and action upon insert', function() {
  spyOn($scope.collection, 'save').and.callThrough();
  
  var doc = {_id: 'new-id'},
    successSpy = jasmine.createSpy('success')
  
  var promise = $scope.collection.save(doc)
  
  spyOn(promise, 'then').and.callThrough();
  
  promise.then(successSpy);
  
  expect(successSpy).toHaveBeenCalledWith({ _id: "test", action: "inserted"});
  $rootScope.$apply();
});
```

There must something I don't understand about how ```save()``` is functioning in the Jasmine environment, so I have not been successful in finding a test strategy to ensure that the return value of this method does not regress.  btw, I have experimented with ```$scope.$apply()```, ```$scope.$digest()```,  ```$timeout.flush()```, and some other methods, but have not yet found why the ```save()``` method does not call-through on its returned promise.

Urigo:
Sorry I haven't dig in your code to much but can you see if the way we are handling promises on the `subscribe` tests helps you maybe?
https://github.com/Urigo/angular-meteor/blob/master/tests/integration/angular-meteor-subscribe-spec.js

jonmc12:
cool, new tests!  I did review, but the subscribe-spec really tests that a promise is returned, vs testing the value of the return.

It is possible to create a similar test, ```spyOn($scope.collection, 'save').and.callFake(...```, but this kind of test just mocks a promise return, and not the value.  ie, using a mock would not have prevented the regression that caused this bug.

PR #715, opened 2015-09-26
(meteorSubscribe) Added options object argument with a user supplied onStop callback

davidyaha:
* User supplied onStop will be called after subscription is already active and was stopped by an error or call to stop method.
 * This does not break the stable API of meteorSubscribe and and adds a way to be notified on an event of subscription stopping after it was resolved.
 * This PR also removes deprecated argument of Meteor.subscribe.

### DIFF
--- a/modules/angular-meteor-subscribe.js
+++ b/modules/angular-meteor-subscribe.js
@@ -8,7 +8,6 @@ angularMeteorSubscribe.service('$meteorSubscribe', ['$q',
     this._subscribe = function(scope, deferred, args) {
       var subscription = null;
       var lastArg = args[args.length - 1];
-      var promiseFulfilled = deferred.promise.$$state.status;
 
       // User supplied onStop callback
       // save it for later use and remove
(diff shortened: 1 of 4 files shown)


Urigo:
@davidyaha very nice!  can you please add tests for the scenarios and also link here to the deprecated argument and the onStop function docs?

davidyaha:
Sure :) 
Meteor Docs only describe the new callback and do not mention the old one but in the code you can see that it deprecated: https://github.com/meteor/meteor/blob/master/packages/ddp-client/livedata_connection.js#L466-L496

The onStop docs is described vaguely in here: http://docs.meteor.com/#/full/meteor_subscribe
It is a bit of a mess so here is a quote:
"The onStop callback is called with a Meteor.Error if the subscription fails or is terminated by the server. If the subscription is stopped by calling stop on the subscription handle or inside the publication, onStop is called with no arguments."

Urigo:
great thank you.
About the tests, there are right now exactly 4 failing tests that are waiting for a pull request in Meteor itself to be merged and shipped, so make sure that when you add your tests, this is still the situation.

davidyaha:
OK cool.

davidyaha:
https://github.com/Urigo/angular-meteor/pull/718

davidyaha:
Opened a PR to master instead: https://github.com/Urigo/angular-meteor/pull/718

PR #718, opened 2015-09-27
(meteorSubscribe) Added options object argument with a user supplied onStop callback 

davidyaha:
* User supplied onStop will be called after subscription is already active and was stopped by an error or call to stop method.
* This does not break the stable API of meteorSubscribe and and adds a way to be notified on an event of subscription stopping after it was resolved.
* Added tests for meteorSubscribe service.
* This PR also removes deprecated argument of Meteor.subscribe. https://github.com/meteor/meteor/blob/master/packages/ddp-client/livedata_connection.js#L466-L496
* Meteor docs quote describing onStop: "The onStop callback is called with a Meteor.Error if the subscription fails or is terminated by the server. If the subscription is stopped by calling stop on the subscription handle or inside the publication, onStop is called with no arguments."

### DIFF
--- a/.docs/angular-meteor/client/content/api/api.subscribe.html
+++ b/.docs/angular-meteor/client/content/api/api.subscribe.html
@@ -47,6 +47,14 @@ Calling $scope.$meteorSubscribe will automatically stop the subscription when th
   <td><p>Optional arguments passed to publisher function on server.</p></td>
   <td><a href="" class="label type-hint type-hint-object">No</a></td>
 </tr>
+<tr>
+  <td>onStop</td>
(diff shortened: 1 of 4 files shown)


Urigo:
@davidyaha is there something we will need to change or add to the docs?
http://angular-meteor.com/api/subscribe

davidyaha:
Yes. I can go ahead and do that as well if you would like.

Urigo:
@davidyaha I would like :+1:

Urigo:
very nice job!
and with the new Meteor 1.2.0.2 all tests passed!

Issue #734, opened 2015-10-02
$meteor.call causing ReferenceError on server


jonmc12:
I've been using angular@1.0.1, and today received an error when using ```$meteor.call```.  This error does not appear when calling ```Meteor.call``` from either the client or the server.

```
Exception while invoking method 'createSpreadSheet' ReferenceError: path is not defined
   at [object Object].Meteor.methods.createSpreadSheet (server/tools.js:147:1)
   at maybeAuditArgumentChecks (livedata_server.js:1692:12)
   at livedata_server.js:708:19
   at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
   at livedata_server.js:706:40
   at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
   at livedata_server.js:704:46
   at tryCallTwo (/Users/jonmc12/.meteor/packages/promise/.0.5.0.h2jsml++os+web.browser+web.cordova/npm/node_modules/meteor-promise/node_modules/promise/lib/core.js:45:5)
at doResolve (/Users/jonmc12/.meteor/packages/promise/.0.5.0.h2jsml++os+web.browser+web.cordova/npm/node_modules/meteor-promise/node_modules/promise/lib/core.js:171:13)
at new Promise (/Users/jonmc12/.meteor/packages/promise/.0.5.0.h2jsml++os+web.browser+web.cordova/npm/node_modules/meteor-promise/node_modules/promise/lib/core.js:65:3)
```

I have resolved the error locally by changing API usage in [package.js](https://github.com/Urigo/angular-meteor/blob/master/package.js#L33).  It works when I change

```
api.use('check');
```
to
```
api.use('check', 'client');
```

Urigo:
Are you using the `check` package or function somewhere?

Urigo:
In anyway we use `check` in the code just on the client side so I'm pushing a fix.

jonmc12:
Thank for fix!  I wasn't sure if needed on server or would have provided pull request

jonmc12:
@Urigo I'm really having trouble with this bug.  I've tried various things, and every time it is fixed for a short period of time and then re-appears.

Yesterday, I fixed by removing the use of the check method in the angular-meteor method and other places in my code.  However, this morning the error reappeared.

Today, the error seems to have temporarily disappeared as I ran ```meteor remove check```.  However, this fix is not working on my colleague's machine.

[...]

PR #743, opened 2015-10-06
updating version of check dependency 1.0.2 -> 1.0.6

jonmc12:
I was still seeing the issue #734.  I'm not exactly certain what is going on, but my best guess is that angular-meteor and my other packages are requesting to use different versions of the ```check``` package as a dependency.

Upgrading to latest version of ```check``` resolved the issue for me.  Hopefully this is final fix needed to resolve this bug.

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -22,7 +22,7 @@
     ],
     [
       "check",
-      "1.0.2"
+      "1.0.6"
     ],
     [
       "dburles:mongo-collection-instances",


jonmc12:
fyi, even with this update, bug #734 still exists.  Probably can't hurt to upgrade to latest version.

Anyhow, I'm unclear of actually how to resolve the bug.  In the meantime, I've commented out my use of the ```check``` method on the server as a work-around.

Urigo:
Closed, details in here:
https://github.com/Urigo/angular-meteor/issues/734#issuecomment-147125851

PR #754, opened 2015-10-12
Feature/meteor1.2 babel

pbastowski:
> Please remove the `blaze-html-templates` package, otherwise you'll get errors about two plugins trying to use the same `.html` extension

- Added Babel support to the script compiler plugin, as per pbastowski:angular-babel. The plugin first compiles the ES6 code with Babel and then imediately runs ng-annotate on the compiled code.
- Bumped package version to 1.2.0-alpha-1


Urigo:
Awesome @pbastowski !

As per @netanelgilad suggestion today on the [strategy doc](https://docs.google.com/document/d/1oHYuuJ-_2epBVuOJ8qJm2Rk30Xm-wA0xhAYw4qfdcLI/edit), maybe it's better to imply a separate package, like in [React and Jsx](https://github.com/meteor/react-packages/blob/devel/packages/react/package.js#L17)?
Second, the right way of versioning is `1.2.0-alpha.1`

pbastowski:
Sure, no problem. I like it. 

If I understand it correctly, instead of including a babel-ng-annotate plugin in the `angular` package, we will imply a new package, say, `angular-babel`, which itself will implement the plugin, Right?

I've fixed the alpha naming in the PR.

... but, it would see that I need to create a new package called `angular-babel`.

Urigo:
yes that's the idea.
Thank you for that

pbastowski:
I have removed the JS plugin support from `angular-meteor` and extracted it into it's own `angular-babel` package https://github.com/pbastowski/angular-babel. The angular-meteor `package.js` in this PR implies `angular-babel`, thus implementing the JS compiler plugin I removed in the first place.

For testing you'll need my `feature/meteor1.2 babel` angular-meteor branch and also clone my `angular-babel` repo into your packages folder.

.npm/plugin/compileNGScript/npm-shrinkwrap.json
```diff
@@ -1,16 +1,433 @@
 {
   "dependencies": {
+    "babel-core": {
```
Urigo: @pbastowski this came from previous changes or from the imply?

lib/runtime.js
```diff
@@ -0,0 +1,564 @@
...
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * https://raw.github.com/facebook/regenerator/master/LICENSE file. An
+ * additional grant of patent rights can be found in the PATENTS file in
+ * the same directory.
+ */
+
+!(function(global) {
```
Urigo: We still need this after the imply?

package.js
```diff
@@ -20,15 +20,16 @@ Package.registerBuildPlugin({
...
-Package.registerBuildPlugin({
-  name: 'compileNGScript',
-  sources: [
-    'plugin/ng-script-compiler.js'
-  ],
-  npmDependencies: {
-    'ng-annotate': '0.15.4'
-  }
-});
+//Package.registerBuildPlugin({
```
Urigo: Delete all of it

package.js
```diff
@@ -60,6 +63,11 @@ Package.onUse(function (api) {
     // Lib Files
     'lib/diff-array.js',
     'lib/get-updates.js',
+
+    //// Babel files
+    //'lib/core-js-no-number.js',
```
Urigo: Also please delete the commented code

plugin/ng-script-compiler.js
```diff
@@ -1,21 +1,73 @@
...
 var processFile = function(file) {
-  var annotated = ngAnnotate(file.getContentsAsString(), {
+
+  var source = file.getContentsAsString();
+  var outputFile = file.getPathInPackage();
+  var to5output = "";
+
+  console.log('Babel compiling: ', outputFile);
+
+  try {
```
Urigo: Shouldn't we imply that configuration as well?

pbastowski:
Uri, I'm thinking if the `imply` is a good thing to do. 

My concern is that once we do it, then others who might want to use TypeScript, for example, will unnecessarily have to also `imply` (include) Babel runtimes in their FE code.

What's your take on that?

pbastowski:
And, yeah, I have to cleanup the files :) I wanted first to hear back from you about my `imply` question above.

Urigo:
@pbastowski yes that's a good point.
Maybe another package named `angular-bundle` or `angular-pack` that implies both `angular` and `angular-babel` can be a good idea.
What do you think?
@netanelgilad

pbastowski:
My internal feeling says, let the people decide which packages they want to add. 

However, for many people and especially new-comers to angular-meteor, it will make a difference if we have an all encompassing bundle, which is what you said you wanted to provide, right? 

So, I vote for `angular-bundle`, which implies `angular`, `angular-babel` and 'angular2-now`.

Urigo:
Yes I think I'm with you on that.

Urigo:
But leave it open for a while, I want to think a bit more about it

Urigo:
@pbastowski did you update `angular-babel` to use regular `.js` files?
I want to release a new version and test it out

Urigo:
@pbastowski @netanelgilad @DAB0mB I vote against `angular-bundle` for now.
The developer will add `angular` and `angular-babel` and remove `ecmascript`.
That's easy enough.
Now let's just see that the `angular-babel` is ready, test with `socially` and release

pbastowski:
@Urigo No, I did not. I was waiting for you to get back to me.

Good decision! Let the people add the packages they want. So, here's what I will do:

1) I will add in-memory caching for the compiler to speed things up considerably.

2) I propose to make one change to Babels generated ES5 code. I want to output SystemJS bundled ES5 modules instead of simple ES5 code. This will allow people to do 

```javascript
import { Feature2 } from 'client/Feature2/Feature2'
```

The above lazy-executes the imported `Feature2`. See my [`pbastowski:typescript`](https://github.com/pbastowski/meteor-typescript) package for another example.

I need to actually speak with you about this tonight and to vote on yes or now, if possible, as this is a potentially breaking feature, which would require that we modify the `Socially` app accordingly. But, it would also move us into the realm of ES6 modules and SystemJS.

BTW: My `typescript` package caches in-memory and also includes my SystemJS package. Tests on my own apps, where I used this package instead of Babel, show that everything works just fine. So, what I'm saying is that we could use TypeScript instead of Babel for most cases. No type annotations are required in any of my code and everything works, so, it's a viable alternative to Babel. And if you want to use types then you can.

Urigo:
@pbastowski I would prefer to wait with adding SystemJS as ES6 modules is really close to get into Meteor (version 1.3) - https://github.com/meteor/meteor/pull/5475
and I would want to be based on the Meteor solution and not another

About caching - I think it's important, but not enough to wait with the release, how long to you think it will take you to add and test it ?

pbastowski:
I'm on it now. Hopefully in an hour or so it will all be working.

pbastowski:
@Urigo It's done. However, I don't have rights to publish a root package called `angular-babel`. It is a root package, because it is not prefixed with `pbastowski:` So, do I prefix it or do I get rights to publish a root package or how do we handle the publishing on Atmosphere?

For testing, just clone this repo into your test project's packages folder and then

    meteor add angular-babel

I used it in my socially app with `angular2-now` and it worked as expected.

Urigo:
Let's leave it under your name for now as the official ES6 is different

pbastowski:
I already have a package called `pbastowski:angular-babel` for Meteor 1.1, so, this new one for Meteor 1.2 I will call `pbastowski:ng-babel`.

Urigo:
Won't it be easier to do a major version bump with the same package?

pbastowski:
Sure, I'd love to, but... the code is too different. How do I maintain two very different code bases in the same package for two different versions of Meteor?

I don't know how to do this. Maybe you can help me with this?

pbastowski:
Oh, I think I see what you meant. To bump the major package version and then publish the new Meteor 1.2 code to Atmosphere under the same package name. This will of course only run in Meteor 1.2, but that's the plan. Right?

Urigo:
Yes, then you can keep publish versions for the older versions, like Angular is doing

pbastowski:
ok, done.

The `pbastowski:angular-babel` package is ready :)

Once again, I tested it with my `angular2-now` version of Socially and it works as expected.

PR #759, opened 2015-10-14
Update tutorial.step_09.md

Erdou:
Next chapter's intro is wrong: it's saying we're going to learn filters instead of the deployment.


Urigo:
Thanks!

Issue #772, opened 2015-10-18
Bug with upserting docs with ObjectIds


nickbenes:
I noticed a problem while going through the Update and Remove step of the meteor.com [Meteor-Angular Tutorial](https://www.meteor.com/tutorials/angular/update-and-remove). I kept seeing 
```
Error: Meteor requires document _id fields to be non-empty strings or ObjectIDs
```
in the console, and tracked it down to a step during `AngularMeteorCollection._upsertDoc` where it calls `stripDollarPrefixedKeys`. During this step, a document with `ObjectID("12345") for its `_id` value would have that converted to the plain JavaScript Object `{_str: "12345"}`, which is neither a string nor an ObjectID.

It looks like `angular-meteor` always saves objects with string `_id`s, which bypasses this problem, but it comes up if a document is created from the `meteor mongo` console and then updated, and could also arise from documents created by another library. Here's an example, starting after [Step 3: Collections](https://www.meteor.com/tutorials/angular/collections) where the controller code includes
```
$scope.tasks = $meteor.collection(Tasks);
```
From a mongo console run
```
db.tasks.insert({ "text": "foo", "createdAt": new Date() })
```
Then from the Chrome browser console select an element within the controller (so `$scope` becomes available in the console) and run
```
$scope.tasks[0].text = "bar";
$scope.$apply();
```
where the index 0 is replaced with the actual index of the "foo" task in your array.

Submitting PR shortly to fix this. I tried to construct an integration test to go along with it, but I don't think I can produce an ObjectId within the current framework. Definitely open to suggestions there. Thanks for taking a look.

PR #773, opened 2015-10-18
Fix ObjectID being converted to Object during updates

nickbenes:
Meteor collections currently require that _id be either a string or the EJSON
version of Mongo's ObjectID. When documents are being saved with ObjectID (such
as in early steps of the current angular tutorial on meteor.com), the
`stripDollarPrefixedKeys` method that is called by the `_upsertDoc` method of
`AngularMeteorCollection` causes ObjectIDs to be converted into POJOs, which
causes Meteor to throw errors about their data types and messes everything up.

This commit causes values with EJSON type 'oid' (such as the server-side
ObjectIDs) to pass through `stripDollarPrefixedKeys` without modification.

### DIFF
--- a/modules/angular-meteor-utils.js
+++ b/modules/angular-meteor-utils.js
@@ -29,6 +29,7 @@ angularMeteorUtils.service('$meteorUtils', [
       if( !angular.isObject(data) ||
         data instanceof Date ||
         data instanceof File ||
+        EJSON.toJSONValue(data).$type === 'oid' ||
         (typeof FS === 'object' && data instanceof FS.File)) {
         return data;
       }


nickbenes:
Fixes #772

Urigo:
@nickbenes thank you so much for that!

About testing, can you please add simple unit tests for `stripDollarPrefixedKeys`?
about the particular error test, maybe create an stripDollarPrefixedKeys manually with this type of id?

Urigo:
@nickbenes I'm merging your change, but can you please add tests in another pull request?

nickbenes:
@Urigo Definitely, and sorry for missing your comment earlier this week.

jacobdr:
I feel like there have been a few issues now where the `stripDollarPrefixedKeys` method has led to some errors. 

Maybe we ought to dig into the [test suite](https://github.com/Urigo/angular-meteor/blob/7febfed399a360fd906999044b775262eb1b2c9c/tests/integration/angular-meteor-utils-spec.js#L73) and make sure it is robust enough..

Urigo:
@jacobdr would love some help with that!

PR #785, opened 2015-10-22
Updates with $push for new subarrays (just kidding)

MarkPhillips7:
Discovered that updates were storing what should be an array as an object. 
Added test and got the test to pass.
Also fixed a couple of tests that were failing only because the expectation was set incorrectly I believe.


Urigo:
Thanks @MarkPhillips7 !!

PR #786, opened 2015-10-23
Update api.subscribe.html

EyalRonel:
Updated example to include the the actual "declaration" for the Books collection.
Updated comments within the example


Urigo:
Thanks @EyalRonel good catch

PR #798, opened 2015-10-29
Update ionic.step_06.md

dinesh36:
Need to "add" the meteor package for publish-composite not "remove" the package


Urigo:
good point :)

PR #802, opened 2015-10-30
Update api.getReactively.html

dajocarter:
Example for $scope.getCollectionReactively had a dollar sign in front of the method.


Urigo:
thanks!

PR #809, opened 2015-10-31
Update tutorial.step_21.md

Tallyb:
Editorial review of the step's tutorial. 

Question: you are using $stateProvider but did not explicitly define ui-router as dependency. How does that work?


Urigo:
Thanks @Tallyb , I think it's a question if we want the dependencies on the package level of the app level.
I guess the package level is better, but it is also harder to maintain maybe....

Tallyb:
@Urigo Well, at the minimum, I think it worth noting the decision and trade-offs in the tutorial. WDYT?

PR #810, opened 2015-10-31
fix typos in angular2 tutorial in steps 0, 2, 3

DumpOfTheVar:
Step 0 - Bootstrapping
After the first run, you will find tge -> After the first run, you will find the

Step 2 - Dynamic Template
"party of parties" -> "#party of parties"

Step 3 - 3-Way data binding
that should happen reactively happen reactively -> that should happen reactively.

Remove lines "Now every change what happens to the this.parties variable should automatically be saved to the local client-side minimongo DB and synced to the server-side MongoDB and all the other clients in realtime!". If I understand right the changes from minimongo will affect this.parties but changes in this.parties will not affect data in minimongo.


Urigo:
Thanks @DumpOfTheVar

PR #832, opened 2015-11-06
Update tutorial.step_03.md

AlmightyR:
Fixed two instances where the 'app.ng.js' file was (still) being referred to as 'app.js' (only used up to tutorial step 2.4).


Urigo:
@AlmightyR thank you for that!
but we are about to release `angular@1.2.0` which doesn't require `ng.js` anymore...

PR #835, opened 2015-11-06
Fixed: `app.ng.js` file being called `app.js`

AlmightyR:
Fix to two instances where the `app.ng.js` was being called `app.js` outside the code snippets.

The `app.js` file was renamed to `app.ng.js` at tutorial step 2.4.


Urigo:
@AlmightyR thank you for that!
but we are about to release `angular@1.2.0` which doesn't require `ng.js` anymore...

PR #836, opened 2015-11-06
Fix collection changes were being notified with no updates

DAB0mB:


tests/integration/angular-meteor-diff-array-spec.js
```diff
@@ -101,6 +103,25 @@ describe('diffArray module', function() {
...
+        {_id: "b", outer: {inner: "bar"}}
+      ];
+
+      diffArray(oldCollection, newCollection, {
+        addedAt: addedAtSpy,
+        removedAt: removedAtSpy,
+        changedAt: changedAtSpy
+      });
+
+      expect(addedAtSpy).not.toHaveBeenCalledWith();
```
idanwe: Wasn't it should be `toHaveBeenCalled()` otherwise the test can be pass if it the callback call with something that it is not empty arguments?

idanwe: @Urigo @DAB0mB I tested it and indeed it need to be `.not.toHaveBeenCalled()`

Urigo: and then the test still passes?
can you submit a pull request?

idanwe: Already open :) #837

PR #837, opened 2015-11-06
Fix assert .not.toHaveBeenCalled()

idanwe:
As described here https://github.com/Urigo/angular-meteor/pull/836/files#r44182491


Urigo:
Thanks @idanwe

PR #838, opened 2015-11-06
Correct test spec to check if callbacks have been called

DAB0mB:


Urigo:
@DAB0mB isn't that what @idanwe has already done?

DAB0mB:
Yes I've just seen it, I will close it

PR #839, opened 2015-11-07
Mention how to add "account-password" package

dinesh36:
I think somewhere in description it should be mentioned to how to add the accounts-password package
"meteor add accounts-password"


Urigo:
@dinesh36 it's not necessarily `accounts-password`, there are many accounts package like `accounts-sms`..
that's why I linked to the Meteor docs.
maybe we can change that to something more obvious like you meant, but less specific?

dinesh36:
Agreed :+1:

Issue #806, opened 2015-10-31
An issue in angular2 tutorial step3


dai-shi:
http://angular-meteor.com/tutorials/angular2/3-way-data-binding

The example does not work until we add the `main.ts` file because the server side collection is not loaded, probably.

barbatus:
I guess the example means downloaded zip archive?
If yes then each link at the top point to the beginning of the current state, so doesn't contain code below.
If you want complete code go to the next step and download.

dai-shi:
No, I meant the tutorial itself. Sorry for the confusion.

I went through the tutorial carefully from the beginning without downloading the archive,
and at step 3 it didn't work as explained, specifically when inserting data from the console.

Has anyone experienced this before?

barbatus:
Ok, now I see. You are saying that it didn't work at Inserting Parties from the Console.
Yeah, i guess paragraph says to create main.ts and import collection should go before.

dai-shi:
Yeah, thanks for helping the clarification. At first, I tried to "improve the doc", and found out that it was not trivial. I would like someone to take part in it.

aaronksaunders:
I still cannot get this code to work... even after downloading the code from the tutorial page... any guidance would be appreciated

Urigo:
I've just deployed @barbatus recent changes, please let me know if it's working for you now

dai-shi:
Cool. Thanks! @barbatus @Urigo

PR #842, opened 2015-11-08
Ng2 Tutorial Updates

barbatus:
1. Rearranges substeps of the 3rd step to define collection on the server side before inserts from the mongo terminal (#806, [#7](https://github.com/Urigo/meteor-angular2.0-socially/issues/7))
2. Fixes some minor issues: global zone replaced to the specific Angular 2 one, explanation for 'angular2/router' added


barbatus:
Deployed - http://ng2-tutorial.meteor.com//

Issue #847, opened 2015-11-09
Angular 2 Step 3 app.ts status


kimberger:
When following the tutorial the `app.ts` file is different to what the documentation is suggesting it should be at that stage.

Urigo:
Merged and deployed
thank you @kimberger

PR #848, opened 2015-11-09
Fix step3 docs

kimberger:
Having followed the Angular 2 tutorial the `app.ts` file differs to what the documentation suggests it should be at this stage. Issue #847


Urigo:
@barbatus

barbatus:
ok

PR #858, opened 2015-11-10
Fix "main" path in bower.json

canac:
The `main` field in bower.json refers to a non-existent directory. This prevents tools like [main-bower-files](https://www.npmjs.com/package/main-bower-files) from working correctly. This PR fixes that field to point to the main script in the `dist/` directory.


Urigo:
@idanwe

PR #863, opened 2015-11-11
Remove unnecessary return statement in object saving

DAB0mB:


Urigo:
@DAB0mB can you add a test to see that a promise is returned?

PR #921, opened 2015-12-09
Correct deprecated messages

sebakerckhof:
Correct deprecated messages


sebakerckhof:
On a side note, how are we to replace these in our apps using 1.3?

Issue #939, opened 2015-12-11
Should properties be reactive too ?


hlehmann:
I want to load data depending on a property of my object

with 1.2

    $scope.data = loadData() //$scope.$meteorObject(...)
    $scope.$watch('data.category', category=> loadCategoryDetails(category))

with 1.3

    $scope.helpers({
      data() {
        return loadData() //findOne ...
      }
      category() {
        return loadCategoryDetails($scope.data.category)
      }
    })

`$scope.data = {category: '1'}` will refresh category

`$scope.data.category = '1'` will not

Deadly0:
@hlehmann looks strange. Did you tried use deep watch (send to `$watch` 3rd param `true`) or use $watchCollection instead?

hlehmann:
well it works well using $scope.$watch but not with the helper.
The question is more : is it appropriate to use the helpers when it is combine with angular changes (like in a form)

sebakerckhof:
+1

Deadly0:
@hlehmann it should work with form. If you want to watch helper use `this.autorun(() => {/* your helper here */})` It's works for me.

canac:
I am also having trouble getting changes in my scope properties to update my helpers. @Deadly0, could you give a more complete example of using autorun? Because `$scope.autorun(() => $scope.helpers({ ... }) );` is still not working for me. Is that what you meant?

Deadly0:
@canac `selectedId` should be helper

```
const app = angular.module('test', ['angular-meteor']);

app.controller('TestCtrl', function($scope) {
  $scope.selectItem = function(item) {
    $scope.selectedId = item._id;
  };

  $scope.helpers({
    selectedId: null,
    items() {
      return Items.find();
    },

    selectedItem() {
      return Items.findOne($scope.selectedId);
    },
  });
});
```

canac:
@Deadly0 Yes, that works, both in that contrived example and in my actual application! Thank you so much for your help!

Regarding @hlehmann's original question, I found that helpers are not deeply reactive. For example, given the following helpers:

```js
$scope.helpers({
  selectedId: { id: null },
  items() {
    return Items.find();
  },
  selectedItem() {
    return Items.findOne($scope.selectedId.id);
  },
});
```

[...]

PR #954, opened 2015-12-13
(feature) Split reactiveProps and helpers into different methods

dotansimha:
@Urigo 

#939


sebakerckhof:
So now we're going to deep watch all objects, what does this mean for performance compared to version 1.2?

dotansimha:
@sebakerckhof , it still does not the full and final solution. we are still working on this issue, and we might find another solution for the deep watch.

Urigo:
@sebakerckhof you are definitely right, please look at the discussion here: https://github.com/Urigo/angular-meteor/issues/955

PR #971, opened 2015-12-18
feat (subscribe):  make subscribe return a the same interface as Mete…

netanelgilad:
…or.subscribe.
@Urigo

### DIFF
--- a/packages/angular-meteor-data/modules/angular-meteor-reactive-scope.js
+++ b/packages/angular-meteor-data/modules/angular-meteor-reactive-scope.js
@@ -13,15 +13,22 @@ angularMeteorReactiveScope.run(['$rootScope', '$reactive', '$parse', function ($
     }
 
     subscribe (name, fn = angular.noop, resultCb) {
-      return this.autorun(() => {
+      let result = {};
+      let autorunComp = this.autorun(() => {
         let args = fn() || [];
-
         if (!angular.isArray(args)) {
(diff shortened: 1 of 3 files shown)


packages/angular-meteor-data/modules/angular-meteor-reactive-scope.js
```diff
@@ -13,15 +13,22 @@ angularMeteorReactiveScope.run(['$rootScope', '$reactive', '$parse', function ($
...
         }
 
-        this.stopOnDestroy(Meteor.subscribe(name, ...args, resultCb));
+        let subscriptionResult = Meteor.subscribe(name, ...args, resultCb);
+        this.stopOnDestroy(subscriptionResult);
+        result.ready = subscriptionResult.ready.bind(subscriptionResult);
+        result.subscriptionId  = subscriptionResult.subscriptionId;
       });
+
+      result.stop = autorunComp.stop.bind(autorunComp);
```
Urigo: @netanelgilad will that stop only the autorun or also the subscription?

netanelgilad: `If you call Meteor.subscribe within a reactive computation, for example using Tracker.autorun, the subscription will automatically be cancelled when the computation is invalidated or stopped; it's not necessary to call stop on subscriptions made from inside autorun`. :smile:

Urigo: Awesome!

PR #1038, opened 2015-12-29
angular-meteor-auth.bundle.js doesn't pass minification

ThomsC:
I suppose now it does.


Urigo:
@idanwe

idanwe:
@Urigo @ThomsC The fix should be at the [source code](https://github.com/Urigo/angular-meteor/blob/master/packages/angular-meteor-auth/modules/auth.js#L3)
and then create new bundle

Urigo:
@ThomsC do you want to create another PR?

ThomsC:
Ok sure I'll do it

PR #1049, opened 2015-12-31
I though medium blog is not up to date enough!

pacozaa:
angular-meteor have been updated and the first place I go to for read about the update is the obvious "Blog" menu on the top then I found it isn't up to date. So I think you should change "Blog" url as my suggestion or up date your medium blog.

What do you think?


Urigo:
Cool :)

PR #1062, opened 2016-01-04
changed anguluar-meteor.com urls to use http since https not supported

jboothe:
To eliminate frustration for new visitors, fixed the angular-meteor.com urls to point to http instead of https. The https calls are not resolving.


Urigo:
Great!  and thanks for the email addition.
I hope that soon I will have time to get a nice https certificate

PR #1089, opened 2016-01-11
Fix broken link to Meteor Blog

canac:


ShMcK:
Good catch!

@Urigo this looks good :)

Urigo:
I will add the `s` :)

Tallyb:
Does that mean previous articles are gone? ;)

canac:
@ShMcK and @Urigo, I tried HTTPS first but got a privacy error in Chrome because the certificate belongs to a248.e.akamai.net, so I went with HTTP instead.

PR #1165, opened 2016-01-30
Add reactivateCollection()

DAB0mB:


Urigo:
@canac let me know if that works

packages/angular-meteor-data/tests/integration/reactive.spec.js
```diff
@@ -388,5 +460,35 @@ describe('angular-meteor.reactive', function() {
...
+      it('should return model', function() {
+        vm.myProp = 10;
+        expect(vm.$reactivateCollection('myProp')).toEqual(10);
+      });
+
+      it ('should register a scope collection watcher', function() {
+        vm.myProp = 'myProp';
+        var watchSpy = spyOn(scope, '$watchCollection');
+
+        vm.$reactivateCollection('myProp', false);
```
canac: `$reactivateCollection` does not have an `isDeep` argument, so this `false` parameter is extraneous.

packages/angular-meteor-data/tests/integration/reactive.spec.js
```diff
@@ -388,5 +460,35 @@ describe('angular-meteor.reactive', function() {
...
+        scope = $rootScope.$new();
+        vm = scope.$viewModel({});
+      });
+
+      afterEach(function() {
+        scope.$destroy();
+      });
+
+      it('should return model', function() {
+        vm.myProp = 10;
```
canac: This should probably be an array because `$watchCollection` expects a collection, not an object.

canac:
@Urigo It looks pretty good! I added a few suggestions about the tests, but everything else looks great!

Urigo:
@canac great comments, would you be willing to PR the branch to add your changes?

PR #1172, opened 2016-02-01
bithound fixes

ozsay:


ozsay:
@Urigo you can either merge this or wait for the 1.3.6 merge, They both contain there fixes

PR #1216, opened 2016-02-08
1.3.6

ozsay:


packages/angular-meteor-data/modules/core.js
```diff
@@ -0,0 +1,104 @@
+angular.module('angular-meteor.core', [
+  'angular-meteor.utils',
```
kamilkisiela: @ozsay Shouldn't be here `angular-meteor.utilities`?

packages/angular-meteor-data/modules/view-model.js
```diff
@@ -0,0 +1,90 @@
+angular.module('angular-meteor.view-model', [
+  'angular-meteor.utils',
```
kamilkisiela: Same as here
https://github.com/Urigo/angular-meteor/pull/1216/files#r52520790

ozsay: Yeah it should

Thanks

kamilkisiela: I'm going through angular-meteor line by line so it came up. Good job @ozsay with 1.3.6! :)

Issue #644, opened 2015-09-01
angular-meteor conflicts with meteor-job-collection when included after


mxab:
Hi,

I just noticed that the angular-meteor package in combination with the meteor job collection throws an error if included in the wrong order
```
'The global definition of Mongo.Collection has changed since the job-collection package was loaded. Please ensure that any packages that redefine Mongo.Collection are loaded before job-collection.'
```

I guess this is a related issue
https://github.com/vsivsi/meteor-job-collection/issues/58

I cannot really see where some Mongo.Collection shadowing is happening, maybe in one of the dependencies does it

Not sure if there is anything that can be done, but I thought I mention it

mxab:
A very simple workaround for now is to make sure that the angular packages is listed before the job collection in the _packages_ file

```
angular
...
vsivsi:job-collection
```

Urigo:
@mxab can you check if that happens by one of the package dependencies?

Urigo:
I think I had the same thing with his files package..
I'm closing unless someone would like to investigate this package

lekev:
Hi I have the same issue, It seems that angular package use [dburles:mongo-collection-instances](https://github.com/dburles/mongo-collection-instances) as depency
`api.use('dburles:mongo-collection-instances@0.3.4', 'client'); // For getCollectionByName `
[PackageFile](https://github.com/Urigo/angular-meteor/blob/7febfed399a360fd906999044b775262eb1b2c9c/package.js#L45)

the workaround to list angular before the job collection in the packages file and to add 'Mongo.Collection.prototype.constructor = Mongo.Collection;' before the new new JobCollection( ... ) doesn't fix the issue

ozsay:
@Urigo The new API doesn't use this package. It's just for legacy code. Should we do something?

DrDanRyan:
FYI, this also causes issues with `vsivsi:file-collection` which is a very nice package for people wanting to use gridFS and not take on CollectionFS. It is also a very well maintained package compared with CollectionFS.

PR #1232, opened 2016-02-14
Remove mongo collection instance package

lekev:
Remove no needed package mongo-collection-instance . It also fix the issue https://github.com/Urigo/angular-meteor/issues/644

### DIFF
--- a/packages/angular-meteor-data/modules/angular-meteor-collection.js
+++ b/packages/angular-meteor-data/modules/angular-meteor-collection.js
@@ -5,382 +5,377 @@
 'use strict';
 
 var angularMeteorCollection = angular.module('angular-meteor.collection',
-  ['angular-meteor.stopper', 'angular-meteor.subscribe', 'angular-meteor.utils', 'diffArray']);
+    ['angular-meteor.stopper', 'angular-meteor.subscribe', 'angular-meteor.utils', 'diffArray']);
 
 // The reason angular meteor collection is a factory function and not something
 // that inherit from array comes from here:
 // http://perfectionkills.com/how-ecmascript-5-still-does-not-allow-to-subclass-an-array/
 // We went with the direct extensions approach.
 angularMeteorCollection.factory('AngularMeteorCollection', [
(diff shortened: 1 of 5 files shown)


Urigo:
@lekev you are right. 
The question is, does that change makes the `angular` package depends on a new version of Meteor?
Which means we should change that - https://github.com/Urigo/angular-meteor/blob/master/packages/angular/package.js#L10
and if so, to what version and do we really want that?
maybe we do, I'm not sure.  what do you think?

lekev:
the changes is not dependant of the new meteor version,  Since the AngularMeteorCollection is deprecated and replace by the helpers function, we will not use anymore the mongo-collection-instance . 
Actually we still use it [here](https://github.com/Urigo/angular-meteor/blob/496b1d00cbf2db9fa2df71b30d3eeae1f4c5489d/packages/angular-meteor-data/modules/angular-meteor-collection.js#L41) . So maybe should I find a way to replace $meteorUtils.getCollectionByName to not use it, until you release the version 1.4.0

lekev:
I am already use my fork with meteor 1.2.1 , and it works fine

packages/angular-meteor-data/modules/angular-meteor-collection.js
```diff
@@ -34,12 +34,7 @@ angularMeteorCollection.factory('AngularMeteorCollection', [
       // (autorun for cursor.fetch).
       data._hDataAutorun = null;
 
-      if (angular.isDefined(collection)) {
-        data.$$collection = collection;
-      } else {
-        var cursor = curDefFunc();
-        data.$$collection = $meteorUtils.getCollectionByName(cursor.collection.name);
-      }
+      data.$$collection = collection;
```
Urigo: @lekev The thing is that we need this `if` statement so you could define both a cursor or a string.
How do you handle defining a Cursor from a string here?

Urigo:
@lekev In 1.3 I still want and need to support AngularMeteorCollection.
For that I have to be able to get the instance of the Collection by it's string name.
your code would make AngularMeteorCollection not usable.  have you tested the 1.2 API?

PR #1278, opened 2016-03-17
fix(viewmodel): View model and scope can both use 'getReactively'

DAB0mB:

### DIFF
--- a/dist/angular-meteor.js
+++ b/dist/angular-meteor.js
@@ -1539,6 +1539,11 @@
 	    return obj instanceof $rootScope.constructor;
 	  };
 
+	  // Checks if an object is a view model
+	  this.isViewModel = function (obj) {
+	    return _.isObject(obj) && obj.$$dependencies;
+	  };
+
 	  // Checks if two objects are siblings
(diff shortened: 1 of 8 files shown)


kamilkisiela:
What if someone wants to use `$scope.helpers` and `vm.helpers`?
It won't work so should we support it?

```javascript
$scope.helpers({
    foo() { return FooCollection.find({}); }
});

vm.helpers({
    bar() { return BarCollection.find({}); }
});

// check foo
$scope.foo // undefined
vm.foo // defined

// check bar
$scope.bar // undefined
vm.bar // defined
```

Urigo:
@kamilkisiela we are going to solve this scenario as well, even though it's a horrible pattern :)
but because it's a breaking change we will solve it.

DAB0mB:
@kamilkisiela @Urigo See https://github.com/Urigo/angular-meteor/pull/1278/commits/c48a1f552d9f1ba7ffc014713a258676aa02b583#diff-a3547eb36ef3a55c1cedddba5195e00fR392

PR #1288, opened 2016-03-21
Update docs: running locally

kamilkisiela:
`npm link` and `npm run watch` is amazing to use :smiley:

Cotributor won't have to recompile `angular-meteor` or rerun meteor app.

The same thing with symlink and `npm run watch:dev` in Meteor 1.2.

---

Closes: #1288


kamilkisiela:
@ozsay @Urigo 

What do you think?

Urigo:
@kamilkisiela awesome!

PR #1306, opened 2016-03-30
fix(autorun): destroy event listener once autorun has been stopped

DAB0mB:
Once a computation has been stopped the destroy event listener will be removed
to avoid memory leaks and unexpected behaviors

### DIFF
--- a/src/modules/core.js
+++ b/src/modules/core.js
@@ -34,7 +34,8 @@ angular.module(name, [
       }
 
       const computation = Tracker.autorun(fn, options);
-      this.$$autoStop(computation);
+      // Reset to a function that will also stop the listener we just added
+      computation.stop = this.$$autoStop(computation);
       return computation;
     };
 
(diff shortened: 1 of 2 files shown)


Urigo:
@DAB0mB does that mean that the temporary fix [here](https://github.com/Urigo/angular-meteor-auth/issues/8) should be removed?

DAB0mB:
No, it fixes the memory leak issue, however I need to make some adjustments to angular-meteor-auth. Actually it's pretty much done, it's just that I've encountered some problems with the tests which I'm not sure are even related to my fix, but I'm working on it and it should be uploaded in a separate pull request to angular-meteor-auth's repository.

Urigo:
ok Please also have a look at a few new issues opened on `angular-meteor-auth` that are related to `1.0.0`, see if they related

DAB0mB:
I'll be sure to check these.

Urigo:
@DAB0mB can you remove your changes from the `dist` folder?
I guess you didn't want to commit those right?

PR #1318, opened 2016-04-07
add Trello

Hongbo-Miao:
add Trello


Urigo:
The Trello is not updated.  I prefer to track progress here: https://github.com/Urigo/angular-meteor/milestones
I do need to update it, I will update it tomorrow

Issue #1092, opened 2016-01-12
Act on subscription update


sebakerckhof:
let's say I have a chat message component (directive) which takes a message id, so something like `<chat-message message-id="vm.messageId"></chat-message>`

The controller subscribes to the message data:
```js
this.subscribe('chat.messageDetails',()=>[this.getReactively('messageId')]);
```

Now I'd like to have a variable that indicates whether the data is loading, how could I do this?

I can use the `onReady` function to know when the subscription returned and set `this.loading` to false. But how do I know when the subscription is being called, is there an equivalent `onStart`/`onCalled` or something?

sebakerckhof:
Yeah, that's using the onReady function I mentioned, but I need something when the subscription is being called (to set `this.animation` to true)

mgrivera:
Hi @sebakerckhof, 

I have never used *traker.autorun* but ... isn't it just to do something when a *reactive source* changes state? 

This is from angular-meteor's doc: 
```javascript
myModule.controller('MyCtrl', ['$scope', '$reactive', function($scope, $reactive) {
  $reactive(this).attach($scope);

  this.myVar = 10;

  this.autorun(() => {
    console.log('Autorun!!', this.getReactively('myVar'));
  });

  this.myVar = 50; // This will cause the autorun function method to run again
}]);
```

*this.autorun* is just a wrapper of metor's *traker.autorun*. 
 
This is from meteor's doc: 
```javascript
Tracker.autorun(function (c) {
  if (! Session.equals("shouldAlert", true))
    return;

  c.stop();
  alert("Oh no!");
});
```

Could you just check and see if this helps you? ...

Deadly0:
@sebakerckhof i use this way to display loadbar
```
var deferred = $q.defer();
$scope.subscribe('collection',
	() => {
		deferred = $q.defer();
		$scope.dataLoading = deferred.promise; // Loading initialized
		return []
	}, () => {
		deferred.resolve(); // Loading finished
	}
);
```

Urigo:
@sebakerckhof why not setting loading to true one line before the subscription is called?

sebakerckhof:
@Urigo Because the subscription is reactive, and I want a loading indicator everytime it re-runs.
Of course I could do an autorun on the reactive subscription parameters, but I need to do this quite often, so it becomes cumbersome.

kamilkisiela:
[...]

PR #1347, opened 2016-04-28
feat(Core): onStart callback

kamilkisiela:
@Urigo 

Related to #1092 

---

```js
this.subscribe('books', () => [ this.getReactively('published') ], {
  onStart() {}, // new callback
  onReady() {},
  onStop() {}
});
```

---

I used object as a hook container instead of a single variable for `onStart`, so we'll be ready to add new hooks. 

Let's say we want to create `beforeStart` and `afterStart`, it's now a two lines thing.

### DIFF
--- a/src/modules/core.js
+++ b/src/modules/core.js
@@ -45,6 +45,12 @@ angular.module(name, [
       fn = this.$bindToContext($Mixer.caller, fn || angular.noop);
       cb = cb ? this.$bindToContext($Mixer.caller, cb) : angular.noop;
 
+      // Additional callbacks specific for this library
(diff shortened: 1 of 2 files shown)


Urigo:
@kamilkisiela that's very interesting.
Do you think it can also go as a PR to Meteor itself?

kamilkisiela:
@Urigo Yes, it could be done but is it necessary? `Meteor.subscribe()` is a synchronous function so if everything after it would behave the same as in `onStart` callback.

What do you think, Uri?

Urigo:
but if someone uses it separately on the client side?

Issue #1228, opened 2016-02-12
Remove duplicate html-compilers of angular-templates and angular-blaze-templates


Urigo:
packages/angular-blaze-templates-compiler/plugin/ng-html-scanner.js

packages/angular-templates/plugin/ng-html-scanner.js

PR #1348, opened 2016-04-29
[WIP] chore(templates): Move common things to angular-templates-runtime

kamilkisiela:
In addition to #1228 

This creates a new package named `angular-templates-runtime`.

@Urigo why do we keep `cheerio`? I couldn't find any usage of that.


Urigo:
because it was nicer to handle dom that way when scanning.
but maybe there were a lot of iterations since then...
@DAB0mB what do you think about removing `cheerio`?

DAB0mB:
@Urigo There is no need in `cheerio`, as you said it is used to manipulate the DOM but we don't use it anymore

PR #1355, opened 2016-05-11
feat(templates): optional error and warning of missing template

kamilkisiela:
## API

`$angularTemplatesSettings` constant as a plain object with two properties:

- `(boolean) error` - throws an error - *default: true*
- `(boolean) warning` - uses console.warn - *default: true*

Of course `error` overwrites `console.warn`.

## Use cases

If someone wants to avoid throwing errors he can simply set `error` to false and still he will see helpful warnings about missing template.

In different case, when user doesn't want to be notified, he can set both options to false.

### DIFF
--- a/packages/angular-templates/templates-handler.js
+++ b/packages/angular-templates/templates-handler.js
@@ -14,24 +14,36 @@ angular.module('angular-templates', []).config([
   function ($provide) {
     var templatesFileExtension = ['html', 'tpl', 'tmpl', 'template', 'view'];
 
-    $provide.decorator('$templateCache', ['$delegate', function($delegate) {
-      var originalGet = $delegate.get;
+    $provide.decorator('$templateCache', ['$delegate', '$angularTemplatesSettings',
+      function($delegate, $angularTemplatesSettings) {
+        var originalGet = $delegate.get;
 
-      $delegate.get = function(templatePath) {
-        var originalResult = originalGet(templatePath);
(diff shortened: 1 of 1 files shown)


Urigo:
@kamilkisiela can you also add a note to the `msg` about how to use those settings in case you see the warning and want to change it?

kamilkisiela:
@Urigo Sure! :+1:

kamilkisiela:
@Urigo 

Error:

> [angular-meteor][err][404] path/to/template.html - HTML template does not exists! You can disable this **error** by setting $angularTemplatesSettings.**error** to false.

Warning:

> [angular-meteor][err][404] path/to/template.html - HTML template does not exists! You can disable this **warning** by setting $angularTemplatesSettings.**warning** to false.

Urigo:
@kamilkisiela can you also add a link to the docs and PR the docs for it?
Here: http://www.angular-meteor.com/api/1.3.6/settings
(and update to 1.3.11)

kamilkisiela:
@Urigo There is a `$angularTemplatesSettings` because `angular-templates` is independent from `angular-meteor`.

I will work on docs now.

kamilkisiela:
@Urigo Done (Urigo/angular-meteor-docs#92)

> [angular-meteor][err][404] path/to/file.html - HTML template does not exists! You can disable this error by following this guide http://www.angular-meteor.com/api/1.3.11/templates

PR #1356, opened 2016-05-11
refactor(templates): use exports.default with __esModule

kamilkisiela:
```js
#
# Outdated
#
// before
import templateUrl from './template.html';
// after
import { templateUrl } from './template.html';
```

---

Restored previous behavior to export templateUrl by default and added space for additional variables.

### DIFF
--- a/packages/angular-templates-runtime/ng-template-compiler.js
+++ b/packages/angular-templates-runtime/ng-template-compiler.js
@@ -103,12 +103,14 @@ var minifyHtml = function(html) {
 
 function wrapAngularTemplate(templateUrl, contents) {
   return `
+  var templateUrl = "${templateUrl}";
   angular.module('angular-templates')
     .run(['$templateCache', function($templateCache) {
-      $templateCache.put('${templateUrl}', ${JSON.stringify(contents)});
+      $templateCache.put(templateUrl, ${JSON.stringify(contents)});
     }]);
-  if (typeof module !== 'undefined' && module.exports) {
-    module.exports = "${templateUrl}";
+  if (typeof exports !== 'undefined') {
+    exports.__esModule = true;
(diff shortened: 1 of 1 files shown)


trajano:
Can't we have both?  So we do not have to do

    import { templateUrl as template1 } from './template1.html';
    import { templateUrl as template2 } from './template2.html';

kamilkisiela:
@trajano There is a bigger picture of doing it this way but I reverted previous behavior.

@Urigo We can export an additional variable beside a default one.

packages/angular-templates-runtime/ng-template-compiler.js
```diff
@@ -103,12 +103,14 @@ var minifyHtml = function(html) {
...
   return `
+  var templateUrl = "${templateUrl}";
   angular.module('angular-templates')
     .run(['$templateCache', function($templateCache) {
-      $templateCache.put('${templateUrl}', ${JSON.stringify(contents)});
+      $templateCache.put(templateUrl, ${JSON.stringify(contents)});
     }]);
-  if (typeof module !== 'undefined' && module.exports) {
-    module.exports = "${templateUrl}";
+  if (typeof exports !== 'undefined') {
```
trajano: since we're using angular can this be changed to `angular.isDefined`

kamilkisiela: It could but I see no point

trajano: Just consistency when using AngularJS I guess.

trajano: And it would be nice if we can get this released soon.  It's such a nice feature and will eliminate a lot of repetition.

PR #1462, opened 2016-10-28
Fix issue in the style compiler for the deployment

barbatus:
https://github.com/Urigo/angular2-meteor/issues/427


dotansimha:
@barbatus Also updated the compilers package for Blaze

PR #1570, opened 2017-02-21
removed type as per tslint

irongollem:
Linting fails on this since the assignment already suggests it's of type number. Makes the whole build fail sillily enough :P All I did was remove the type on this assignment to make tsLint happy.


irongollem:
Sorry for fouling it with 4 commits of such simple nature, but bitHound was generating the issues one at a time sadly.

Urigo:
fixed in https://github.com/Urigo/angular-meteor/pull/1644
sorry @irongollem for not merging yours....

Issue #1577, opened 2017-03-02
this.getReactively not triggering autorun


sebakerckhof:
I have a variable that updates and is being called inside an autorun function.
If I watch the variable with `$scope.$watch` I see it changes. However, it's not triggering the autorun function.

Example reproduction:
https://github.com/sebakerckhof/ngmeteor-bug

If you run the application, you'll see a list. Clicking on the list updates an element in minimongo, which in turns changes a variable. The variable is passed to another component, which should trigger an update, but it doesn't.

E.g. this line is being called: https://github.com/sebakerckhof/ngmeteor-bug/blob/master/imports/ui/foo-details/foo-details.component.js#L18

But not this one: https://github.com/sebakerckhof/ngmeteor-bug/blob/master/imports/ui/foo-details/foo-details.component.js#L23

I know I could probably create the application in a different way, but this is a minimal reproduction of a more complex scenario where this flow is justified.

ping @Urigo

PR #1592, opened 2017-04-18
fix(#1577): Prevent thi…

sebakerckhof:
Fixes: https://github.com/Urigo/angular-meteor/issues/1577

An autorun function can trigger components being compiled and thus other autorun functions in those component controllers to be ran. Since this happens while the first autorun is running, these autorun functions think they are nested in the first autorun function.

As a result, when the parent autorun is invalidated, these child autorun functions are stopped and do not re-run in these scenarios. 

This change is probably the easiest way to fixing this. A better way might be to only run a digest cycle after the autorun function completes, but this would require a much bigger code change.

A side effect from this change is that if you actually want to have nested autoruns, you have to do it like this:

```js
this.autorun(() => {
 Tracker.autorun(() => {});
});
```

Instead of
```js
this.autorun(() => {
  this.autorun(() => {});
});
```

### DIFF
--- a/src/modules/core.js
+++ b/src/modules/core.js
@@ -131,7 +131,12 @@ angular.module(name, [
         !this.$$phase &&
         !this.$root.$$phase;
 
-      if (isDigestable) this.$digest();
+      if (isDigestable) {
+        // If a digest cycle in one autorun triggers another autorun,
+        // we want to run this second autorun in a non-reactive manner.
+        // thus preventing inner autoruns from being dependent on their parents.
+        Tracker.nonreactive(() => this.$digest());
+      }
     };
 
     // Creates a promise only that the digestion cycle will be called at its fulfillment
--- a/tests/client/core.spec.js
+++ b/tests/client/core.spec.js
@@ -90,6 +90,20 @@ describe('angular-meteor.core', function() {
 
         expect(stop.calledOnce).to.be.true;
       });
+
+      it('should run digest in a non-reactive manner, so autoruns triggered by the digest are not dependent on other autoruns', function() {
+        scope.$watch('foo',function(val) {
+          if (val) {
+            var computation = scope.autorun(angular.noop);
+            expect(computation._parent).to.be.null;
+          }
+        });
+
+        scope.autorun(function() {
+          scope.foo = 'baz';
+        });
+       
+      });
     });
 
     describe('subscribe()', function() {


DAB0mB:
An autorun function can trigger components being compiled

Where is that? Can you give me a reference as a piece of evidence? The closest reference I've found was [this](https://github.com/sebakerckhof/angular-meteor/blob/4bee3083efa4d2c02da32dc6b198c4d78fa818f1/src/modules/reactive.js#L145), but you can see that right after there's a non-reactive function scope being executed, so I'm a bit confused.

It seems like you managed to find the source for that problem, however your PR seems like one of these solutions which might break other stuff when trying to run nested computations:

```js
this.autorun(() => {
 Tracker.autorun(() => {});
});
```

There are lots of `angular-meteor` applications out there which might stop working because of it, that's why I'm worried, thus, I'd rather have the full robust solution implemented than a temporary workaround. After we agree on a proper solution I think it would be also necessary to have some tests implemented for such a scenario.

sebakerckhof:
@DAB0mB it can be seen in this repo:
https://github.com/sebakerckhof/ngmeteor-bug

To the best of my understanding, here's what happens:
This autorun function: https://github.com/sebakerckhof/ngmeteor-bug/blob/master/imports/ui/foo-list/foo-list.component.js#L25
Sets the 'activeFooId'.

This triggers this `foo-details` component to be rendered (as per the `ng-if` statement): https://github.com/sebakerckhof/ngmeteor-bug/blob/master/imports/ui/foo-list/foo-list.component.html#L8

As a result this autorun in the controller of that `foo-details` component: https://github.com/sebakerckhof/ngmeteor-bug/blob/master/imports/ui/foo-details/foo-details.component.js#L22

The problem is that this autorun thinks it is nested inside the first autorun. Therefore if the first autorun is invalidated (e.g. when this query is updated: https://github.com/sebakerckhof/ngmeteor-bug/blob/master/imports/ui/foo-list/foo-list.component.js#L21 ), it stops the second autorun from running. That's because `tracker` stops nested autoruns from running, since they should automatically re-run if they're nested and still required. But this is not the case here since they're not actually nested.

DAB0mB:
@sebakerckhof I think a better solution would be putting the `$digest()` method in a non-reactive scope: https://github.com/sebakerckhof/angular-meteor/blob/4bee3083efa4d2c02da32dc6b198c4d78fa818f1/src/modules/core.js#L139

So we would have something like:

```js
    // Digests scope only if there is no phase at the moment
    $$Core.$$throttledDigest = function() {
      const isDigestable = !this.$$destroyed &&
        !this.$$phase &&
        !this.$root.$$phase;

      if (isDigestable) Tracker.nonReactive(() => {
        this.$digest();
      }
    };
```

This way whenever we launch the digestion cycle, created components would run in a non-reactive manner, thus preventing inner autoruns from being dependent on their parents. What do you think?

sebakerckhof:
Yeah, I also thought about adding it there, but was afraid of possible other impact since `$$throttledDigest` is called from a lot of places and I don't feel familiar enough with the code. But I guess as long as you don't have autoruns nested inside watchers nested inside other autoruns (which there's probably never a reason for) this should be fine.

DAB0mB:
So I like that solution better actually, because this way you can still run computations inside another. Can you please change your PR accordingly? And check if all the tests pass and add a new one which checks that scenario.

sebakerckhof:
Yeah, it's the better solution. 
I changed this PR to go with your solution, I added a test and verified all tests pass.

DAB0mB:
Thanks @sebakerckhof. @Urigo I can confirm this. Can you merge this one and release a new patch?

PR #1644, opened 2017-09-21
Port to Angular 4

darkbasic:
I ported angular-meteor to Angular 4. I tested this patch on a big project which otherwise couldn't upgrade and it works for me. The code may not be perfect, so please review.


uncanard:
Hello guys, 
Do you think it is safe to use the darkbasic:master branch ?
Is a merge planned for Angular 4 compatibility ? Thanks

darkbasic:
We plan to eventually merge it, but further testing is welcome. Can you please let us know if it works for your use case?

PR #1655, opened 2017-10-19
Trigger digest when Subscription ready changes

SimonSimCity:
Sorry, but I wasn't really able to wrap my head around how to test this feature ...

In my example, I wanted to call the `ready()` method of a subscription directly in the HTML template of my AngularJS app. This change now makes this possible by running the $digest cycle every time the ready-property of a subscription (and thereby also the response of the `ready()` method) changes.

My first attempt included a call of `ready()`, but this broke an existing test, where you rely on calling the `ready()` method. Therefore I changed it to use the meteor-internal property `_subscriptions` to directly access the ready-computation for this subscription.

The autorun I added is registered as a dependency of the outer autorun and will be stopped once the outer autorun is called again.

Looking forward to your response. I'll already start using this in my project.


SimonSimCity:
@Urigo any update on this one? I'm using this for quite a while now and don't see any downsides. If there are, or things I overlooked, I'd like to know.

Also if it's simply because of the fact that you don't want to put any effort into Meteor for AngularJS but rather focus on it's successor.

SimonSimCity:
I've updated the code to use the public method, provided by Meteor instead some internal properties, that aren't part of the public interface of Meteor.

PR #1666, opened 2017-10-20
chore: update renovate.json

rarkins:
The previous config is not correct, it would need to be `{"meteor":{"enabled":true}}` and not `{"meteor":true}. However using `extends` is simpler so I submit the PR this way.


Urigo:
thank you @rarkins !

PR #1710, opened 2017-10-26
Fix tests exit codes for CI

darkbasic:
Travis was actually broken since we moved all the commands in a separate script (`run_tests.sh`), such a way the exit codes from mocha were actually hidden behind the bash script and Travis couldn't manage to pick them up. I made them propagate up to Travis.
Because of this we didn't notice that tests did actually break despite the "green flag" from Travis.
To fix them please merge https://github.com/Urigo/angular-meteor/pull/1706 together with this commit.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -8,6 +8,7 @@ cd angular-compilers-tests
 rm -rf node_modules
 meteor npm install
 meteor npm test
+exit_code=$?; if [ ${exit_code} -gt 0 ]; then exit ${exit_code}; fi
 cd ../../examples/angularcli-meteor
 npm install
 npm run meteor-client:bundle
@@ -15,3 +16,4 @@ npm run api:reset
 export CHROME_BIN=chromium-browser
 npm run test
 concurrently "npm run start" "npm run api" "sleep 90; npm run e2e" --kill-others --success first
+exit_code=$?; if [ ${exit_code} -gt 0 ]; then exit ${exit_code}; fi


darkbasic:
@rarkins our CI was broken because our bash script didn't propagate exit codes properly up to Travis. Now it's fixed, but we need to rerun all the tests before being able to merge the renovate PRs, what should we do? Should we just wait for renovate to rerun all the tests (how do we know when it happened?) or it's better to simply close all the PRs and wait for them to be re-created? Thanks

rarkins:
@darkbasic technically Renovate doesn't *run* any tests itself, it just creates PRs that trigger CI like any other PR. And until Renovate sees a need to update the branch again (e.g newer version of dependency released, or branch becomes conflicted with master and needs recreating), then it won't touch them and hence they won't run.

One Renovate configuration option you can add is `"rebaseStalePrs": true`, this has the effect of telling Renovate to rebase its PRs any time master has been updated. i.e. it should trigger Renovate to now rebase all existing PRs which should then cause tests to run. So you could add that configuration option and then revert it later if you prefer not to have so much "noise" from all the Travis rebuilds it will typically cause.

rarkins:
See https://github.com/Urigo/angular-meteor/pull/1712 in case you want this option.

FYI if you were to *close* PRs manually you would want to rename them first (e.g. append a `-` to PR title) so that Renovate doesn't think you're deliberately ignoring them (such as because of bad patch).

rarkins:
One more thing: if you have already edited any of these Renovate branches then Renovate won't rebase them in case it risks losing data you've added. In such a case then you should rename then close, to get a fresh replacement created.

darkbasic:
Thanks @rarkins you've been very clear, concise and helpful. I merged your PR.

PR #1725, opened 2017-11-02
Angular dependencies updated to 5.0.0 Stable

ardatan:
Meteor 1.6 and Angular 5 released, then Angular Universal doesn't work correctly due to conflict between Zone.js and Fibers.
This PR updates examples, compiler packages and tests.


examples/MeteorCLI/all-in-one/imports/polyfills.ts
```diff
@@ -1,5 +1,14 @@
-import 'zone.js';
 import 'core-js/es7/reflect';
+const oldThen = Promise.prototype.then;
+Object.defineProperty(Promise.prototype, "then", {
```
Urigo: @ardatan could you add a comment pointing out the right issues and PRs on zone.js and Meteor?

PR #1942, opened 2018-08-08
Update commands on README.md

tiagomsmagalhaes:
Followed commands and symlinks weren't working because thought it would work with entering `api/`


ardatan:
Thank you!

PR #1965, opened 2019-05-01
Add support for namedExports when using Rollup plugin

araad:
Currently there is no support for namedExports when using the Rollup plugin which generates an unusable build when a 3rd party package is required and not resolved during the compilation. With this change we can allow the user to define a file called 'named-exports.json' at the root of the project and define all the namedExports required in the code.

This file could contain something like this:
```
{ './module.js': ['foo', 'bar' ] }
```

If the file does not exist, the rollup method will be passed the null value as is the current behavior.


ardatan:
Thank you @araad !

PR #1975, opened 2020-01-19
Update compiler dependencies to most current versions

dohomi:
This is a quick fix to pin the required atmosphere packages to the most recent versions. Upgrade to Meteor 1.9 is otherwise not possible

### DIFF
--- a/atmosphere-packages/angular-compilers/package.js
+++ b/atmosphere-packages/angular-compilers/package.js
@@ -14,9 +14,9 @@ Package.registerBuildPlugin({
   use: [
     // Uses an external packages to get the actual compilers
     'ecmascript@0.10.5',
-    'angular-typescript-compiler@0.3.2',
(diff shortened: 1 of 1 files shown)


fidelsam1992:
Please approve merge to publish the new version of angular-compliers package compatible with Meteor 1.9

SimonSimCity:
Btw ... if you update the meteor development tools, which is what happens automatically, this won't work anymore - at least this was the case for me. I wasn't able to run `meteor update` anymore because of this package - no matter what version of Meteor I was referencing in the configuration. Worse of course when trying to run using Meteor 1.9 or later, where even starting the app didn't work anymore.

This is a show-stopper to me, because it keeps people from using Meteor together with Angular. We are trying to use it, but struggle a lot as we need someone with good knowledge of Meteor to even get people started ...

